PR #7, opened 2021-02-08
Fix webxrmanager

0b5vr:
Original PR I've made: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51033

### Why

The primary motivation is to follow changes made in https://github.com/mrdoob/three.js/pull/20754.

### What

- Update the type definition of WebXRManager.
- Add a test about WebXR VR, `webxr-vr-cube.ts`.
- Update the type definition of XRSession, since the test I mentioned above doesn't work without this change.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
LGTM, thank you 👍🏼

PR #9, opened 2021-02-08
[Feature] re-add mesh generics

### LINKED ISSUE #3: [feature] Make Meshes generic again
I believe we should undo this commit mrdoob/three.js@8c4ff573de466cd97d1074c73b7d8e675a0fe1f5 & make meshes a generic again that would extend BufferGeometry and Material. Not doing so causes loads of headaches using React.useRef() with GLTF models & you end up having to Typecast every time you want to access a property on the material that isn't part of the default Material. Thoughts?

joshuaellis:
### Why

resolves #3

### What

Undone the work done in three's commit – https://github.com/mrdoob/three.js/commit/8c4ff573de466cd97d1074c73b7d8e675a0fe1f5 because you end up having to typecast everything when using `react-three-fiber` and the material type was wrong, it would always complain that it could be an array.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/objects/InstancedMesh.d.ts
+++ b/types/three/src/objects/InstancedMesh.d.ts
@@ -5,8 +5,11 @@ import { Mesh } from './Mesh';
 import { Matrix4 } from './../math/Matrix4';
 import { Color } from './../math/Color';
 
-export class InstancedMesh extends Mesh {
-    constructor(geometry: BufferGeometry, material: Material | Material[], count: number);
+export class InstancedMesh<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
+> extends Mesh {
+    constructor(geometry: Geom, material: Mat, count: number);
 
     count: number;
     instanceColor: null | BufferAttribute;
--- a/types/three/src/objects/InstancedMesh.d.ts
+++ b/types/three/src/objects/InstancedMesh.d.ts
@@ -6,10 +6,10 @@ import { Matrix4 } from './../math/Matrix4';
 import { Color } from './../math/Color';
 
 export class InstancedMesh<
-    Geom extends BufferGeometry = BufferGeometry,
-    Mat extends Material | Material[] = Material
-> extends Mesh {
-    constructor(geometry: Geom, material: Mat, count: number);
+    TGeometry extends BufferGeometry = BufferGeometry,
+    TMaterial extends Material | Material[] = Material | Material[]
+> extends Mesh<TGeometry, TMaterial> {
+    constructor(geometry: TGeometry, material: TMaterial, count: number);
 
     count: number;
     instanceColor: null | BufferAttribute;
--- a/types/three/src/objects/LineLoop.d.ts
+++ b/types/three/src/objects/LineLoop.d.ts
@@ -3,10 +3,10 @@ import { Material } from './../materials/Material';
 import { BufferGeometry } from '../core/BufferGeometry';
 
 export class LineLoop<
-    Geom extends BufferGeometry = BufferGeometry,
-    Mat extends Material | Material[] = Material
-> extends Line {
-    constructor(geometry?: Geom, material?: Mat);
+    TGeometry extends BufferGeometry = BufferGeometry,
+    TMaterial extends Material | Material[] = Material | Material[]
+> extends Line<TGeometry, TMaterial> {
+    constructor(geometry?: TGeometry, material?: TMaterial);
 
     type: 'LineLoop';
     readonly isLineLoop: true;
--- a/types/three/src/objects/LineLoop.d.ts
+++ b/types/three/src/objects/LineLoop.d.ts
@@ -2,8 +2,11 @@ import { Line } from './Line';
 import { Material } from './../materials/Material';
 import { BufferGeometry } from '../core/BufferGeometry';
 
-export class LineLoop extends Line {
-    constructor(geometry?: BufferGeometry, material?: Material | Material[]);
+export class LineLoop<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
+> extends Line {
+    constructor(geometry?: Geom, material?: Mat);
 
     type: 'LineLoop';
     readonly isLineLoop: true;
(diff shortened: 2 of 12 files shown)


types/three/src/objects/InstancedMesh.d.ts
```diff
@@ -5,8 +5,11 @@ import { Mesh } from './Mesh';
 import { Matrix4 } from './../math/Matrix4';
 import { Color } from './../math/Color';
 
-export class InstancedMesh extends Mesh {
-    constructor(geometry: BufferGeometry, material: Material | Material[], count: number);
+export class InstancedMesh<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
```
0b5vr: This should be preferable instead imo?

```suggestion
    Mat extends Material | Material[] = Material | Material[]
```

0b5vr:
I'm not sure the first argument `TGeometry` is no longer required or not, since `Geometry` has been removed since r125... I'm aware that it's a breaking change but adding the removed generics is already a breaking change in either way

types/three/src/objects/InstancedMesh.d.ts
joshuaellis: I think doing that brings about the same problem I'm trying to avoid:

```ts
let mesh: THREE.Mesh

// more code

particleLight = new THREE.Mesh(new THREE.SphereGeometry(4, 8, 8), new THREE.MeshBasicMaterial({ color: 0xffffff }));
particleLight.material.depthTest = true // this throws a TS error because depthTest does not exist on Material[]
scene.add(particleLight);
```

So while it will pick up if someone uses Material[], I don't think it should be a default.

joshuaellis:
So the `TGeometry` generic is not for defining whether it's a `Geometry` or `BufferGeometry`. This is for using different types of `BufferGeometry` where something like [TubeGeometry](https://threejs.org/docs/#api/en/geometries/TubeGeometry) has different parameters you can access compared to `BoxGeometry`.

0b5vr:
Oh they actually have its own members! That makes sense

types/three/src/objects/InstancedMesh.d.ts
0b5vr: Yeah, but I still think the default type parameter should use its broadest definition, as it's hard to assume that just saying `THREE.Mesh` excludes use of multiple materials. Probably a preference / opinionated (and I'm not a professional of type systems 😅).

Methuselah96: @joshuaellis Couldn't you define `particleLight` as `THREE.Mesh<BufferGeometry, Material>`? I agree with @FMS-Cat that it should default to all possible values.

Methuselah96: I'm not a big fan of calling them `Geom` and `Mat` since those are just abbreviations of the original types. Can we stick with the old `TGeometry` and `TMaterial` to make it clear that they are type generics?

types/three/src/objects/InstancedMesh.d.ts
```diff
@@ -5,8 +5,11 @@ import { Mesh } from './Mesh';
...
+> extends Mesh {
```
Methuselah96: We should add the generics each place we're extending from  `Mesh`.

```suggestion
> extends Mesh<TGeometry, TMaterial> {
```

types/three/src/objects/LineLoop.d.ts
```diff
@@ -2,8 +2,11 @@ import { Line } from './Line';
 import { Material } from './../materials/Material';
 import { BufferGeometry } from '../core/BufferGeometry';
 
-export class LineLoop extends Line {
-    constructor(geometry?: BufferGeometry, material?: Material | Material[]);
+export class LineLoop<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
+> extends Line {
```
Methuselah96:
```suggestion
> extends Line<TGeometry, TMaterial> {
```

types/three/src/objects/LineSegments.d.ts
```diff
@@ -11,8 +11,11 @@ export const LineStrip: number;
  */
 export const LinePieces: number;
 
-export class LineSegments extends Line {
-    constructor(geometry?: BufferGeometry, material?: Material | Material[]);
+export class LineSegments<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
+> extends Line {
```
Methuselah96:
```suggestion
> extends Line<TGeometry, TMaterial> {
```

types/three/src/objects/SkinnedMesh.d.ts
```diff
@@ -4,8 +4,11 @@ import { Skeleton } from './Skeleton';
 import { Mesh } from './Mesh';
 import { BufferGeometry } from '../core/BufferGeometry';
 
-export class SkinnedMesh extends Mesh {
-    constructor(geometry?: BufferGeometry, material?: Material | Material[], useVertexTexture?: boolean);
+export class SkinnedMesh<
+    Geom extends BufferGeometry = BufferGeometry,
+    Mat extends Material | Material[] = Material
+> extends Mesh {
```
Methuselah96:
```suggestion
> extends Mesh<TGeometry, TMaterial> {
```

types/three/test/lights/lights-spotlight.ts
```diff
@@ -55,7 +55,7 @@ function init() {
 
     const geometry = new THREE.PlaneGeometry(2000, 2000);
 
-    let mesh = new THREE.Mesh(geometry, material);
+    const mesh = new THREE.Mesh(geometry, material);
```
Methuselah96, line 58: Can you add a new test that tests this change more specifically (i.e., something that would require a cast without generics, but doesn't require a cast with generics)?

Methuselah96 (changes requested):

joshuaellis:
I added a new text for the `materials-envmaps-hdr`, also captures two loaders which is great, I originally converted on master to see the issues without generics & moved them over.

Methuselah96 (dismissed):

Issue #2, opened 2021-02-06
r126


Methuselah96:
This issue exists to track the status of the changes for r126 that might also need type changes.

- [x] ~Material: Move flatShading to supported materials. (https://github.com/mrdoob/three.js/pull/21200)~ https://github.com/three-types/three-ts-types/pull/4
- [x] ~Update Geometry.js (https://github.com/mrdoob/three.js/commit/33497f2fad7afa87f2d052b8b2e84bbe347b5425)~ no change to types in the status quo
- [x] ~Add possibility to bind 3D textures and 2D textures array as color attachment to framebuffer (https://github.com/mrdoob/three.js/pull/20111)~ https://github.com/three-types/three-ts-types/pull/15
- [x] ~Docs: Updated frustumCulled description. (https://github.com/mrdoob/three.js/commit/159b7c0313d451a31e64b36ca8a3d4e4e515e641)~ https://github.com/three-types/three-ts-types/pull/8
- [x] ~Add nodematerial-light (https://github.com/mrdoob/three.js/commit/5e6fa8cc99)~ #21 
- [x] ~Remove ClosedSplineCurve3 from THREELegacy (https://github.com/mrdoob/three.js/commit/771ec7ce67)~ not typed in the status quo
- [x] ~PointLightShadow: Use own private objects (https://github.com/mrdoob/three.js/commit/aa30ab0d24~ no type changes
- [x] ~USDZExporter: Added textures support (https://github.com/mrdoob/three.js/commit/e99aa7ba5b) (https://github.com/mrdoob/three.js/commit/3759ffa3da)~ not typed in status quo
- [x] ~3DMLoader material access (https://github.com/mrdoob/three.js/commit/f290c25901) – not sure about this, need to compare full code with `.d.ts`~ no type changes in status quo
- [x] ~rename ModelViewProjectionMatrixNode -> ModelViewProjectionNode (https://github.com/mrdoob/three.js/commit/8d4da0276c)~ not typed
- [x] ~GLTFLoader: Remove support for MSFT_texture_dds (https://github.com/mrdoob/three.js/commit/c1946d75cb)~ https://github.com/three-types/three-ts-types/pull/22
- [x] ~Add support for request headers to ImageBitmapLoader inGLTFLoader (https://github.com/mrdoob/three.js/commit/2846e2fcc6)~ no type changes to the status quo
[...]

PR #10, opened 2021-02-10
feature (GLTFLoader): Add beforeRoot and afterRoot to GLTFLoaderPlugin

0b5vr:
Related issue: #2 

It's directed to `dev` since it's a change that will be applied in the next release (r126).

### Why

To follow the recent change to the GLTFLoader.

See: https://github.com/mrdoob/three.js/pull/21207

### What

- Add `beforeRoot` and `afterRoot` to the interface `GLTFLoaderPlugin` .
- Add a test `gltfloader-plugin.ts`, an example that uses a user-defined plugin in GLTFLoader.
    - Because of the test, I had to add missing methods of `GLTFParser`.

### Checklist

    - I'm already on there
-   [x] Add a test 
-   [x] Ready to be merged

### Points need review

- I'm not sure I should add those `GLTFParser` methods.
    - Most of them are probably not intended to be used by end developers.
        - I might want to ask about this to contributors of GLTFLoader 🤔 
    - I had to add them because of the test, as I mentioned above.
        - I can do `(parser as any).loadMesh` instead though.
- Are those `{ [key: string]: any }` good enough?
    - They are using the schema of GLTF, I would rather add type definitions of GLTF schema if I have to type them.
- The model of the example is missing.
    - It's working for type checking purpose so I think it's okay to leave it as is
    - The working one written in JS, just in case: https://glitch.com/edit/#!/eight-outstanding-approach

### DIFF
--- a/types/three/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/types/three/examples/jsm/loaders/GLTFLoader.d.ts
@@ -78,6 +78,8 @@ export class GLTFParser {
 }
 
 export interface GLTFLoaderPlugin {
+    beforeRoot?: () => Promise<void> | null;
+    afterRoot?: (result: GLTF) => Promise<void> | null;
     loadMesh?: (meshIndex: number) => Promise<Group | Mesh | SkinnedMesh> | null;
     loadBufferView?: (bufferViewIndex: number) => Promise<ArrayBuffer> | null;
     loadMaterial?: (materialIndex: number) => Promise<Material> | null;
--- a/types/three/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/types/three/examples/jsm/loaders/GLTFLoader.d.ts
@@ -1,10 +1,14 @@
 import {
     AnimationClip,
+    BufferAttribute,
+    BufferGeometry,
     Camera,
     Group,
+    InterleavedBufferAttribute,
     Loader,
     LoadingManager,
     Mesh,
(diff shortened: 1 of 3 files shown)


joshuaellis:
Thanks, @FMS-Cat can you please add a test for this? `test/loaders/*`. I think our motivation going forward is try & slowly implement tests where we can as we edit files.

0b5vr:
@joshuaellis What should I do about GLTF model? I think I can just refer to a missing model file and give a link to the model as a comment instead.

joshuaellis:
Yeah I'm not sure how you should tackle that one, I suggest give it a go? and If it's not working, we can find a solution together when the test is written.

0b5vr:
Added an example. Also had to add `GLTFParser` methods. See the "Points need review" section on the description.

joshuaellis:
> I'm not sure I should add those GLTFParser methods.

I think it's fine tbh, i mean the actual function isn't exposed to the users from what I can see if the THREE Api (might be wrong) but even if it was, it should be defined I suppose?

> Are those { [key: string]: any } good enough?

Again because this on the `GLTFParser` function I think it's okay. Let's see if someone's opinionated enough they can always fix it.

> The model of the example is missing.

It passes the test, i think that's fine tbh. It won't be the only loader we do this with.

Great work & thank you 🎉

joshuaellis (approved):

0b5vr:
thanks!

PR #11, opened 2021-02-10
DataTextureLoader.load() should return DataTexture

marquizzo:
### Why

Looking [at the source code](https://github.com/mrdoob/three.js/blob/350f0a021943d6fa1d039a7c14c303653daa463f/src/loaders/DataTextureLoader.js#L22-L103), and also [the documentation](https://threejs.org/docs/index.html#api/en/loaders/DataTextureLoader.load), `DataTextureLoader.load()` should return a `DataTexture` instead of `void`.

### What

I simply replaced `:void` with `:DataTexture` in the return type for the `.load()` method.

### Checklist
-   [x] Ready to be merged

### DIFF
--- a/types/three/src/loaders/DataTextureLoader.d.ts
+++ b/types/three/src/loaders/DataTextureLoader.d.ts
@@ -10,7 +10,7 @@ export class DataTextureLoader extends Loader {
         onLoad: (dataTexture: DataTexture) => void,
         onProgress?: (event: ProgressEvent) => void,
         onError?: (event: ErrorEvent) => void,
-    ): void;
+    ): DataTexture;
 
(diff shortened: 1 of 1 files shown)


joshuaellis:
Hi @marquizzo would you mind creating a test for this? We normally just copy the relevant html example from the `three` repo.

joshuaellis (approved):
Reviewing this file, it doesn't appear that there's a suffiecent `html` example to take from the `three` repo as it's part of most other loaders e.g. `RGBELoader`. So i'm going to approve this merger based on that.

joshuaellis:
@all-contributors please add @marquizzo  for code

marquizzo:
Thanks @joshuaellis ! I meant to build the test, but work has been really busy. Glad it could be merged without it!

PR #15, opened 2021-02-20
[r126] – webglrendertarget set texture

joshuaellis:
### Why

part of `r126` – mrdoob/three.js#20111

#### WebGLRenderTarget

* added `setTexture` 
* added `depth` as argument to `setSize`
* `constructor` args should be partial

#### fflate.module

* added .d.ts

#### DataTexture2DArray

* `constructor` args should be partial

### Checklist

-   [x] Ready to be merged


types/three/test/renderers/renderers-renderTarget-texture2DArray.ts
```diff
@@ -0,0 +1,243 @@
+import * as THREE from 'three';
```
Methuselah96, line 1: Where is this test file coming from? Is this test file why we're making the parameters in the constructors optional? It may be correct, but I'm suspicious since it's not documented that these parameters are optional, and the code doesn't default those properties to anything.

Methuselah96 (changes requested):

types/three/test/renderers/renderers-renderTarget-texture2DArray.ts
joshuaellis: No sorry, I had a discussion on the `three` repo about the WebGLRenderer optional params, I need to edit this example, but its taken from the repo, it's about testing `setTexture` in `WebGLRenderTarget`

Methuselah96 (approved):

PR #17, opened 2021-02-22
Add back missing properties to Face

ccincotti3:
### Why

Missing types on Face3, referring to https://threejs.org/docs/#api/en/core/Face3

### What

Re-added them back in 

### Checklist

-   [x] Ready to be merged


joshuaellis:
Hi @ccincotti3, I've just looked at the link you sent, and it's a dead link? You're also updating something in `Raycaster` can you please provide a new source for why the `.d.ts` file needs to be changed please?

PR #18, opened 2021-02-22
[126] Add ArrowHelper descriptions.

CodyJasonBennett:
### Why

This pull request addresses [mrdoob/three.js@`4f3d377`](https://github.com/mrdoob/three.js/commit/4f3d3775ac) for r126, as listed in #2.

### What

I've filled in missing defaults and descriptions in the ArrowHelper class and its methods.

Is this enough, or do I have to explicitly define defaults in its constructor and methods [like we did for `ArrowHelper.type`](https://github.com/CodyJasonBennett/three-ts-types/blob/609e51e5a7b84f68217e9dd121aeb0810c86d313/types/three/src/helpers/ArrowHelper.d.ts#L27-L30)?

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/helpers/ArrowHelper.d.ts
+++ b/types/three/src/helpers/ArrowHelper.d.ts
@@ -8,12 +8,12 @@ import { Object3D } from './../core/Object3D';
 
 export class ArrowHelper extends Object3D {
     /**
-     * @param [dir=new THREE.Vector3( 0, 0, 1 )] Direction from origin. Must be a unit vector.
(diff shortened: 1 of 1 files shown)


joshuaellis:
IMO I think it's great you added the comment stuff! I don't think we need to define the defaults in the TS files really....

BUT, it would be really helpful if you could add webgl_shadowmesh as a test file... thinking it could be `test/objects/objects-shadowmesh.ts` 🙏🏼

CodyJasonBennett:
Sure thing. I'll take a look.

Methuselah96 (approved):
LGTM

PR #19, opened 2021-02-22
[r126] add copyTextureToTexture3D to WebGLRenderer

joshuaellis:
### Why

mrdoob/three.js@b5ff213 part of #2

### What

added copyTextureToTexture3D to WebGLRenderer

### Checklist

-   [x] Ready to be merged


types/three/src/renderers/WebGLRenderer.d.ts
```diff
@@ -430,6 +433,14 @@ export class WebGLRenderer implements Renderer {
      */
     copyTextureToTexture(position: Vector2, srcTexture: Texture, dstTexture: Texture, level?: number): void;
 
+    copyTextureToTexture3D(
```
Methuselah96, line 445: Can you add a JSDoc comment for this method?

types/three/src/renderers/WebGLRenderer.d.ts
```diff
@@ -430,6 +433,14 @@ export class WebGLRenderer implements Renderer {
...
+        sourceBox: Box3,
+        position: Vector3,
+        srcTexture: Texture,
+        dstTexture: DataTexture3D,
```
Methuselah96: Can you provide a source for why this should be `DataTexture3D` instead of `Texture`?

Methuselah96 (changes requested):

types/three/src/renderers/WebGLRenderer.d.ts
joshuaellis: I'm going off the example here, plus the name of the function made it seem logical, but I can just put it to `Texture`

Methuselah96: Looks like it supports `DataTexture3D` and `DataTexture2DArray` based on [this](https://github.com/mrdoob/three.js/commit/b5ff213#diff-fc94afb5da943fea9e01a4a16adf82b5433aa6d61af5855f296158a98ed4dcfdR2018), so we can go ahead and narrow it to those types if you want to.

joshuaellis: I've changed this

Methuselah96 (approved):

PR #23, opened 2021-02-23
[r126] Texture encoding

joshuaellis:
### Why

https://github.com/mrdoob/three.js/commit/b11c85898d – part of #2

### What

Remove default for `Texture.encoding`

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/textures/Texture.d.ts
+++ b/types/three/src/textures/Texture.d.ts
@@ -22,7 +22,7 @@ export class Texture extends EventDispatcher {
      * @param [format=THREE.RGBAFormat]
      * @param [type=THREE.UnsignedByteType]
      * @param [anisotropy=1]
-     * @param [encoding=THREE.LinearEncoding]
+     * @param [encoding]
      */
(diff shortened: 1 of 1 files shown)


types/three/src/textures/Texture.d.ts
```diff
@@ -22,7 +22,7 @@ export class Texture extends EventDispatcher {
      * @param [format=THREE.RGBAFormat]
      * @param [type=THREE.UnsignedByteType]
      * @param [anisotropy=1]
-     * @param [encoding=THREE.LinearEncoding]
+     * @param [encoding]
```
Methuselah96, line 25: From what I can tell, `encoding` still has a default from the constructor.

Methuselah96 (changes requested):

types/three/src/textures/Texture.d.ts
joshuaellis: yeah lets get rid of this PR then

PR #24, opened 2021-02-23
[r126] add VOXMesh & VOXDataTexture3D

joshuaellis:
### Why

mrdoob/three.js@c460b28 – part of #2

### What

adds VOXMesh & VOXDataTexture3D

### Checklist

-   [x] Ready to be merged


types/three/examples/jsm/loaders/VOXLoader.d.ts
```diff
@@ -12,3 +12,11 @@ export class VOXLoader extends Loader {
     loadAsync(url: string, onProgress?: (event: ProgressEvent) => void): Promise<object[]>;
     parse(data: ArrayBuffer): object[];
 }
+
+export class VOXMesh extends Mesh {
+    constructor(chunk: object);
+}
+
+export class VOXDataTexture3D extends DataTexture3D {
+    constructor(chunk: object);
```
Methuselah96: Can we either:

1. Properly type `chunk`?
2. Provide a type alias (e.g., `type Chunk = object`) that can be shared between `VOXLoader`, `VOXMesh`, and `VOXDataTexture3D` just to make it clear that they're connected? And then a TODO comment to complete type.

Methuselah96 (changes requested):

types/three/examples/jsm/loaders/VOXLoader.d.ts
joshuaellis: I actually realised I could piece it together using the `load` function, so now its a lot better

Methuselah96 (dismissed):

Methuselah96 (approved):

PR #26, opened 2021-02-24
[r126] add xr estimated light

joshuaellis:
### Why

mrdoob/three.js@a9112d5 – part of #2 

### What

Adds `XREstimatedLight`, `XRHandOculusMeshModel` & `XRHandPrimitiveModel`

### Checklist

-   [x] Ready to be merged


types/three/examples/jsm/webxr/XREstimatedLight.d.ts
```diff
@@ -0,0 +1,32 @@
+import { DirectionalLight, Group, LightProbe, XRFrame, WebGLRenderer, Texture } from '../../../src/Three';
+
+export class SessionLightProbe<TLight extends THREE.Group = THREE.Group> {
+    xrLight: TLight;
```
Methuselah96: Is there a case where this wouldn't be `XREstimatedLight`?

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    renderer: WebGLRenderer;
+    lightProbe: unknown;
```
Methuselah96, line 6: The type for this can be found [here](https://immersive-web.github.io/lighting-estimation/#xrlightprobe-interface) if you want to add it. We should also look into using `@types/webxr` at some point to avoid duplication.

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    xrWebGLBinding: unknown | null;
```
Methuselah96, line 7: The type can be found [here](https://immersive-web.github.io/layers/#XRWebGLBindingtype) if you want to add it.

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    estimationStartCallback: () => void;
+    frameCallback: (time: number, xrFrame: XRFrame) => void;
```
Methuselah96:
```suggestion
    frameCallback: (this: SessionLightProbe, time: number, xrFrame: XRFrame) => void;
```

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(
+        xrLight: TLight,
+        renderer: WebGLRenderer,
+        lightProbe: unknown,
+        environmentEstimation: boolean,
+        estimationStartCallback: () => void,
+    );
+
+    updateReflection: () => void;
```
Methuselah96, line 19: nit: I don't think it makes a difference, but I like to match the syntax between the JS and TS (here and in the rest of the PR). Ignore this comment if you want to.
```suggestion
    updateReflection(): void;
```

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
```diff
@@ -0,0 +1,13 @@
+import { Group, Object3D } from './../../../src/Three';
+
+import { HandModelOptions, XRHandModel } from './XRHandModelFactory';
+
+export class XRHandOculusMeshModel {
+    controller: Group;
```
Methuselah96, line 10: At some point it probably makes sense to have a more specific type than `Group` for `controller` here and elsewhere. You don't have to change it it in this PR, just leaving a note.

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
```diff
@@ -0,0 +1,13 @@
...
+    handModel: XRHandModel;
+    bones: Object3D[];
```
Methuselah96: Looks like these can potentially be `null` (https://github.com/mrdoob/three.js/blob/dev/examples/jsm/webxr/XRHandOculusMeshModel.js#L105).

```suggestion
    bones: Array<Object3D | null>;
```

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    constructor(handModel: XRHandModel, controller: Group, path: string, handedness: any, options: HandModelOptions);
```
Methuselah96: I think `handedness` is either `'left'` or `'right'` based on https://github.com/mrdoob/three.js/blob/dev/examples/jsm/webxr/XRHandOculusMeshModel.js#L15 (for `XRHandPrimitiveModel` as well).

types/three/examples/jsm/webxr/XRHandPrimitiveModel.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Group, Texture } from '../../../src/Three';
+
+import { HandModelOptions, XRHandModel } from './XRHandModelFactory';
+
+export class XRHandPrimitiveModel {
+    controller: Group;
+    handModel: XRHandModel;
+    envMap: Texture | null;
```
Methuselah96, line 12: Is this ever not `null`? I haven't figured out its purpose, but it's probably fine to leave as-is.

types/three/examples/jsm/webxr/XRHandModelFactory.d.ts
```diff
@@ -1,9 +1,17 @@
...
     constructor();
 
-    motionController: any;
+    motionController: XRHandPrimitiveModel | XRHandOculusMeshModel;
+}
+
+export interface HandModelOptions {
+    model?: 'lowpoly';
+    primitive?: 'sphere' | 'box';
 }
```
Methuselah96: I think this should be split two different interfaces:the options for `XRHandOculusMeshModel` and the options for `XRHandPrimitiveModel` since they don't seem to overlap at all. And then the `options` parameter for `createHandModel` should be the options for `XRHandOculusMeshModel`, since the provided options never get passed to `XRHandPrimitiveModel`.

Methuselah96 (changes requested):

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
joshuaellis: I was going to, but if i'm honest, I don't use the XR stuff enough to move it over meaningfully without knowing if it will break something. I will add that type.

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
joshuaellis: I think this is where `@types/webxr` would come in helpful

types/three/examples/jsm/webxr/XREstimatedLight.d.ts
joshuaellis: I wasn't sure, but lets put it to that for now.

joshuaellis:
I've left the unknowns for the time being and opened a separate issue to tackle moving to `webxr` there's no point in me trying to untangle those types when the library really should just be added.

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
```diff
@@ -0,0 +1,23 @@
...
+export class XRHandOculusMeshModel {
+    controller: Group;
+    handModel: XRHandModel;
+    bones: Array<Object3D | null>;
+
+    constructor(
+        handModel: XRHandModel,
+        controller: Group,
+        path: string,
+        handedness: 'left' | 'right',
```
Methuselah96: Can we make this a type (e.g., `type Handedness = 'left' | 'right';`) that can be shared between the two usages of it?

Methuselah96 (changes requested):

types/three/examples/jsm/webxr/XRHandOculusMeshModel.d.ts
joshuaellis: We could, im not sure the best place to put it though? these are `examples` so they're used in isolation...

Methuselah96: Either a new file that can be shared between the two (e.g., `XRHandModelCommon`) or in one of the factory files (e.g., `XRControllerModelFactory` or `XRHandModelFactory`) since it seems like the normal use case is to create a hand model using the factory classes.

joshuaellis: Lets put it in `XRHandModelCommon` I want to avoid *more* files

Methuselah96 (approved):

PR #27, opened 2021-02-24
[r126] Add SSRPass

joshuaellis:
### Why

mrdoob/three.js@05cc046 – part of #2

### What

Adds SSRPass and it's associated files

### Checklist

-   [x] Ready to be merged


types/three/examples/jsm/postprocessing/SSRPass.d.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export class SSRPass extends Pass {
+    width: number;
+    height: number;
+    clear: boolean;
+    renderer: WebGLRenderer;
+    scene: Scene;
+    camera: Camera;
+    groundReflector: Reflector;
```
Methuselah96:
```suggestion
    groundReflector: Reflector | null;
```

types/three/examples/jsm/postprocessing/SSRPass.d.ts
```diff
@@ -0,0 +1,120 @@
...
+    opacity: SSRShader['uniforms']['opacity']['value'];
```
Methuselah96: I'm not sure if looking up the types for these is worth it since it seems unlikely that the type will change and just makes it indirect when it doesn't have to be. Can we just type this as `number` (and update the ones below similarly)?

types/three/examples/jsm/postprocessing/SSRPass.d.ts
```diff
@@ -0,0 +1,120 @@
...
+    output: number;
+    maxDistance: SSRShader['uniforms']['maxDistance']['value'];
+    surfDist: SSRShader['uniforms']['surfDist']['value'];
+    encoding: TextureEncoding;
+    tempColor: Color;
+
+    _selects: Mesh[] | null;
```
Methuselah96: There is a `selects` getter/setter, so we should probably type that instead of `_selects` since users shouldn't be using this property without the getter/setter (along with the other properties that start with an underscore).

types/three/examples/jsm/postprocessing/SSRPass.d.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    constructor(params: SSRPassParams);
+
+    dispose: () => void;
+
+    renderPass: (
+        renderer: WebGLRenderer,
+        passMaterial: Material,
+        renderTarget: WebGLRenderTarget,
+        clearColor: Color | string | number,
```
Methuselah96, line 108: Can we create a type for this (instead of specifying all of them each time)?

types/three/examples/jsm/shaders/SSRShader.d.ts
```diff
@@ -0,0 +1,57 @@
...
+    defines: {
+        MAX_STEP: number;
+        isPerspectiveCamera: boolean;
+        isDistanceAttenuation: boolean;
+        isFresnel: boolean;
+        isInfiniteThick: boolean;
+        isSelective: boolean;
+    };
+    uniforms: {
+        tDiffuse: { value: Texture };
```
Methuselah96: Can we create a type? Maybe something like:
```
type Uniform<TValue> = { value: TValue };
```

types/three/examples/jsm/shaders/SSRShader.d.ts
```diff
@@ -0,0 +1,57 @@
...
+    vertexShader: string;
+    fragmentShader: string;
+}
+
+export interface SSRDepthShader {
+    defines: {
+        PERSPECTIVE_CAMERA: number;
+    };
+    uniforms: {
+        tDepth: { value: Texture };
```
Methuselah96: It should allow `null` if I'm understanding it correctly.
```suggestion
        tDepth: { value: Texture | null };
```

types/three/examples/jsm/shaders/SSRShader.d.ts
```diff
@@ -0,0 +1,57 @@
...
+        cameraNear: { value: number };
+        cameraFar: { value: number };
+    };
+    vertexShader: string;
+    fragmentShader: string;
+}
+
+export interface SSRBlurShader {
+    uniforms: {
+        tDiffuse: { value: Texture };
```
Methuselah96: Here as well:
```suggestion
        tDiffuse: { value: Texture | null };
```

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ReflectorOptions {
+    clipBias: number;
+    textureWidth: number;
+    textureHeight: number;
+    color: number;
+    useDepthTexture: boolean;
+    shader?: ReflectorShader;
+}
+
+export class Reflector<TGeometry extends BufferGeometry = BufferGeometry> extends Mesh {
```
Methuselah96:
```suggestion
export class Reflector<TGeometry extends BufferGeometry = BufferGeometry> extends Mesh<TGeometry> {
```

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    type: string;
```
Methuselah96:
```suggestion
    type: 'Reflector';
```

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    vertexShader: string;
+    fragmentShader: string;
+}
+
+export interface ReflectorOptions {
+    clipBias: number;
+    textureWidth: number;
+    textureHeight: number;
+    color: number;
+    useDepthTexture: boolean;
```
Methuselah96: Shouldn't these all be optional?

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    shader?: ReflectorShader;
+}
+
+export class Reflector<TGeometry extends BufferGeometry = BufferGeometry> extends Mesh {
+    type: string;
+    options: object;
```
Methuselah96:
```suggestion
    options: ReflectorOptions;
```

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    ReflectorShader: ReflectorShader;
+
+    needsUpdate: boolean;
+    maxDistance: number;
+    opacity: number;
+
+    _isDistanceAttenuation: boolean;
+    _isFresnel: boolean;
```
Methuselah96: Remove underscore since they should be accessed using the getter/setter.

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+} from '../../../src/Three';
+
+export interface ReflectorShader {
+    defines: {
+        isDistanceAttenuation: boolean;
+        isFresnel: boolean;
+    };
+    uniforms: {
+        [key: string]: any;
+    };
```
Methuselah96: Can we improve this type? Maybe if we don't care about the specific properties, we can replace `any` with `Uniform<unknown>`.

types/three/examples/jsm/objects/ReflectorForSSRPass.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    color: number;
+    useDepthTexture: boolean;
+    shader?: ReflectorShader;
+}
+
+export class Reflector<TGeometry extends BufferGeometry = BufferGeometry> extends Mesh {
+    type: string;
+    options: object;
+
+    ReflectorShader: ReflectorShader;
```
Methuselah96:
```suggestion
    static ReflectorShader: ReflectorShader;
```

types/three/examples/jsm/shaders/SSRShader.d.ts
```diff
@@ -0,0 +1,57 @@
+import { Matrix4, Vector2, Texture } from '../../../src/Three';
+/**
+ * References:
+ * https://lettier.github.io/3d-game-shaders-for-beginners/screen-space-reflection.html
+ */
+
+export interface SSRShader {
```
Methuselah96, line 8: I believe we actually need to export the objects from this file as well, not just the interfaces.

types/three/examples/jsm/postprocessing/SSRPass.d.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    depthRenderMaterial: ShaderMaterial;
+
+    copyMaterial: ShaderMaterial;
+
+    fsQuad: Pass.FullScreenQuad;
+
+    originalClearColor: Color;
+
+    OUTPUT: {
```
Methuselah96:
```suggestion
    static OUTPUT: {
```

Methuselah96 (changes requested):

types/three/examples/jsm/postprocessing/SSRPass.d.ts
joshuaellis: I think if we're gonna do that, it needs to be a sweeping change, it's a nice idea but a lot of effort..... Maybe we could make a new issue for it?

types/three/examples/jsm/shaders/SSRShader.d.ts
joshuaellis: I'm not sure what you mean?

types/three/examples/jsm/postprocessing/SSRPass.d.ts
Methuselah96: Yeah, that's fine.

types/three/examples/jsm/shaders/SSRShader.d.ts
Methuselah96: Something like:

```
export const SSRShader: SSRShader;
export const SSRDepthShader: SSRDepthShader;
export const SSRBlurShader: SSRBlurShader;
```

joshuaellis:
To do the generic change, i've merged master into this branch, since we're just focussing on r126, I think this should be fine on this opportunity.

Methuselah96:
Can we merge `master` into `dev` instead of this branch, just to make it clear what the changes are in this PR?

joshuaellis:
done. good thinking

types/three/src/renderers/shaders/UniformsLib.d.ts
```diff
@@ -1,3 +1,7 @@
+export interface UniformBuilder<TValue> {
```
Methuselah96: This might be just be my unfamiliarity with WebGL, but when I see `UniformBuilder`, I think of a builder function/class that returns a `Uniform`. Does it not work to use the name `Uniform` or am I missing something? Alternatively we could beef up `IUniform` to have a generic that defaults to `unknown`.

Methuselah96 (changes requested):

types/three/src/renderers/shaders/UniformsLib.d.ts
joshuaellis: I think thats a fair assumption, I wanted to avoid calling it Uniform because there is/was a class/function called Uniform in `three`. However, we could beef up IUniform like so?

```ts
interface IUniform<TValue = any> {
value: TValue
}
```

if you're happy with that? We wanna do a default instead of trying to type all the existing versions of IUniform

Methuselah96: Yeah, that looks good. Perhaps `unknown` would be a safer default?

joshuaellis: yeah, I'm not actually sure the difference between `unknown` and `any`. So I'll take your word and do `unknown`

Methuselah96: Looks like it fails too many tests, so we can just stick with `any` for now. I'll work on making it `unknown` in the future.

joshuaellis: Cool changed 👌🏼

Methuselah96 (approved):

PR #30, opened 2021-02-25
fix: complete optional parameters.

khum08:


joshuaellis:
Thanks for the contribution @khum08, can you provide a source of why this change was made, please?

khum08:
> Thanks for the contribution @khum08, can you provide a source of why this change was made, please?

I just found parameters in those funtion are different to js source file. you means developer don't need to pass 'update' param when they call those function directly?

khum08:
here is js source code.
image
image

joshuaellis (approved):
Looks good to me, thanks for the contribution.

joshuaellis:
@all-contributors please add @khum08  for code

PR #33, opened 2021-02-26
Scene.d.ts: Background does not support WebGLCubeRenderTarget anymore.

Mugen87:
### Why

With `r127`, `Scene.background` does not support `WebGLCubeRenderTarget` anymore, see https://github.com/mrdoob/three.js/pull/21360.

### What

Remove `WebGLCubeRenderTarget` from `Scene.d.ts`.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
Thank you for your contribution!

Mugen87:
For a separate PR: It seems `WebGLRenderTarget` is imported in `Scene.d.ts` although it is never used...

PR #47, opened 2021-03-16
Update Plane.d.ts

Mugen87:
### Why

`Plane.intersectLine()` returns now `null` instead of `undefined`, see https://github.com/mrdoob/three.js/pull/21468.

### What

Replaced `undefined` with `null`.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
Thank you for the contribution ⭐

PR #49, opened 2021-03-17
SAOPass: params.saoBlur is a boolean.

ycw:
### Why

`SAOPassParams.saoBlur` seems to be a boolean according to  https://github.com/mrdoob/three.js/blob/34519f5073c32eca2cf8b571ce03d44a5a80476a/examples/jsm/postprocessing/SAOPass.js#L57
so i suggest that `saoBlur: boolean` instead of `saoBlur: number`.

### Checklist

-   [x] Ready to be merged

Thanks.


joshuaellis (approved):
Thank you for the contribution ⭐

PR #50, opened 2021-03-18
feat: Add auto-completion of built-in shader attribute names

rinsuki:
### Why

I forgot the attribute name of the built-in shaders every time and wasting 3s~1m each time for remember it. 

### What

I added auto-completion to `setAttribute` with String Union. this will save time for everyone (including me) who use this typing little a bit :)

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/core/BufferGeometry.d.ts
+++ b/types/three/src/core/BufferGeometry.d.ts
@@ -6,6 +6,7 @@ import { Vector2 } from './../math/Vector2';
 import { Vector3 } from './../math/Vector3';
 import { EventDispatcher } from './EventDispatcher';
 import { InterleavedBufferAttribute } from './InterleavedBufferAttribute';
+import { BuiltinShaderAttributeName } from '../constants';
 
 /**
  * This is a superefficent class for geometries because it saves all data in buffers.
@@ -91,7 +92,10 @@ export class BufferGeometry extends EventDispatcher {
     getIndex(): BufferAttribute | null;
     setIndex(index: BufferAttribute | number[] | null): BufferGeometry;
 
-    setAttribute(name: string, attribute: BufferAttribute | InterleavedBufferAttribute): BufferGeometry;
+    setAttribute(
+        name: BuiltinShaderAttributeName | (string & {}),
+        attribute: BufferAttribute | InterleavedBufferAttribute,
+    ): BufferGeometry;
     getAttribute(name: string): BufferAttribute | InterleavedBufferAttribute;
     deleteAttribute(name: string): BufferGeometry;
     hasAttribute(name: string): boolean;
--- a/types/three/src/core/BufferGeometry.d.ts
+++ b/types/three/src/core/BufferGeometry.d.ts
@@ -96,9 +96,9 @@ export class BufferGeometry extends EventDispatcher {
         name: BuiltinShaderAttributeName | (string & {}),
         attribute: BufferAttribute | InterleavedBufferAttribute,
     ): BufferGeometry;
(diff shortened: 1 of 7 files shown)


joshuaellis:
Wouldn't this stop people using their own strings for the name of the attribute?

rinsuki:
@joshuaellis no, we can still use custom attribute names.


https://user-images.githubusercontent.com/6533808/111891801-f5b4ca80-8a38-11eb-8ec9-44dac443d6fd.mp4

joshuaellis:
Seems like a nice addition, @Methuselah96  what do you think?

types/three/src/core/BufferGeometry.d.ts
```diff
@@ -91,7 +92,10 @@ export class BufferGeometry extends EventDispatcher {
     getIndex(): BufferAttribute | null;
     setIndex(index: BufferAttribute | number[] | null): BufferGeometry;
 
-    setAttribute(name: string, attribute: BufferAttribute | InterleavedBufferAttribute): BufferGeometry;
+    setAttribute(
+        name: BuiltinShaderAttributeName | (string & {}),
```
joshuaellis, line 96: why is this `string & {}` I didn't think a name could be an object?

types/three/src/core/BufferGeometry.d.ts
```diff
@@ -91,7 +92,10 @@ export class BufferGeometry extends EventDispatcher {
...
+        attribute: BufferAttribute | InterleavedBufferAttribute,
+    ): BufferGeometry;
     getAttribute(name: string): BufferAttribute | InterleavedBufferAttribute;
```
joshuaellis: This should have the same change as above

types/three/src/core/BufferGeometry.d.ts
```diff
@@ -91,7 +92,10 @@ export class BufferGeometry extends EventDispatcher {
...
     deleteAttribute(name: string): BufferGeometry;
     hasAttribute(name: string): boolean;
```
joshuaellis: as should `has` and `delete`

joshuaellis (changes requested):
Can you also add a test that includes the use of `addAttribute` it should live in `test/geometries`

types/three/src/core/BufferGeometry.d.ts
rinsuki: that's hack to prevent conversion from `"const-string" | string` to `string`.

related: https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972

joshuaellis: hm, im not comfortable with this hack, it could bring more problems than solutions – https://github.com/microsoft/TypeScript/issues/29729#issuecomment-706401313

rinsuki: I think those problems are not related with this usecase, but I'm not familiar with those corner cases, so I might have missed something. 

description:
- Type guards don't narrow the variable
  - In `.d.ts` we dont need type guards because `.d.ts` doesn't including implementation.
- No property auto-completion when an object type is in the union
  - it shouldn't occur accept objects as attribute name
- No string auto-completion when extended by generic type
  - thats maybe problem for future, but currently, thats not problem because we don't using generics for attribute name
- Interferes with conditional types that expect `string & {}` to not be a subtype of `object`
  - anyone want to check `name extends object`?

joshuaellis: Okay well please write a test for this feature as requested we can then be sure that the issues above don't relate to what we're doing now.

joshuaellis (approved):
Thanks for the contribution!

PR #59, opened 2021-03-25
Update WebGLRenderer.d.ts

Mugen87:
### Why

`maxMorphTargets` and  `maxMorphNormals` have been removed from the public interface of `WebGLRenderer`, see https://github.com/mrdoob/three.js/pull/21522.

### What

Removes `maxMorphTargets` and  `maxMorphNormals` from `WebGLRenderer.d.ts`.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/renderers/WebGLRenderer.d.ts
+++ b/types/three/src/renderers/WebGLRenderer.d.ts
@@ -197,16 +197,6 @@ export class WebGLRenderer implements Renderer {
      */
     toneMappingExposure: number;
 
-    /**
-     * @default 8
-     */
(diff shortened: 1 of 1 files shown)


joshuaellis:
When doing the `.d.ts` files originally, was there a reason you excluded `private` methods instead of appending them with `private`? Just for my own reference.

Mugen87:
I did not added this code originally so I'm afraid I can answer your question^^. 

But during the time the `d.ts` files resided in the `three.js` repo, private methods or properties were never added to the type declaration files.

joshuaellis (approved):

PR #60, opened 2021-03-26
Update Audio.d.ts

Mugen87:
### Why

The generic `AudioNode` is now used as the type of audio filters, see https://github.com/mrdoob/three.js/pull/21523.

### What

Use `AudioNode` instead of `any` as type for filters.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution ⭐

PR #61, opened 2021-03-28
Fix clone this types

efokschaner:
### Why

Some of three's objects clone themselves via `new this.constructor(...)`. This allows subclasses to be correctly cloned without slicing. By updating the return types to `this`, TypeScript is now aware of this and allows valid code to compile without type assertions.
You can see this in action in the test `types/three/test/materials/materials-envmaps-hdr.ts` where the type assertions are no longer necessary.

### What

Updated the return types to this where appropriate, after auditing the three implementations.

### Checklist

-   [x] Ready to be merged

I also fixed the test command in contributing.md as the existing instructions did not work out of the box for me.

Thanks in advance!

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -115,6 +115,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "efokschaner",
+      "name": "efokschaner",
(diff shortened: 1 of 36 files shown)


joshuaellis (approved):
Awesome work, thanks for the contribution!! ⭐

PR #68, opened 2021-04-04
Update matrix4 methods

Jhuni0123:
### Why

`Matrix4` has methods [`makeShear`](https://threejs.org/docs/#api/en/math/Matrix4.makeShear) ([code](https://github.com/mrdoob/three.js/blob/r127/src/math/Matrix4.js#L676)) and [`setFromMatrix3`](https://threejs.org/docs/#api/en/math/Matrix4.setFromMatrix3) ([code](https://github.com/mrdoob/three.js/blob/r127/src/math/Matrix4.js#L84)), but types missed them.

### What
- Added `Matrix4.makeShear`
- Added `Matrix4.setFromMatrix3`

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -124,6 +124,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "Jhuni0123",
+      "name": "Jonghun Park",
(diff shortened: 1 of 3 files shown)


joshuaellis (approved):
Thanks!

PR #82, opened 2021-05-27
Update constants.d.ts

puxiao:
### Why

https://github.com/mrdoob/three.js/blob/dev/src/constants.js

```
export const MOUSE = { LEFT: 0, MIDDLE: 1, RIGHT: 2, ROTATE: 0, DOLLY: 1, PAN: 2 };
```

<br>

https://github.com/three-types/three-ts-types/blob/master/types/three/src/constants.d.ts

```
export enum MOUSE {
    LEFT,
    MIDDLE,
    RIGHT,
    ROTATE,
    DOLLY,
    PAN,
}
```

<br>

**I think it's necessary to tell the user clearly the value of the enumeration object.**

### DIFF
--- a/types/three/src/constants.d.ts
+++ b/types/three/src/constants.d.ts
@@ -2,12 +2,12 @@ export const REVISION: string;
 
 // https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent.button
 export enum MOUSE {
-    LEFT,
-    MIDDLE,
-    RIGHT,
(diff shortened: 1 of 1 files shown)


joshuaellis:
@all-contributors add @puxiao for code

joshuaellis (approved):

joshuaellis:
Thanks for the contribution!

PR #85, opened 2021-05-27
WebGLMultipleRenderTargets: Add new class.

Mugen87:
### Why

A new type of render target was added with https://github.com/mrdoob/three.js/pull/16390.

### What

Adds type declaration file for `WebGLMultipleRenderTargets`.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -176,6 +176,7 @@ export * from './objects/Group';
 /**
  * Renderers
  */
+export * from './renderers/WebGLMultipleRenderTargets';
 export * from './renderers/WebGLMultisampleRenderTarget';
 export * from './renderers/WebGLCubeRenderTarget';
 export * from './renderers/WebGLRenderTarget';
--- a/types/three/src/renderers/WebGLMultipleRenderTargets.d.ts
+++ b/types/three/src/renderers/WebGLMultipleRenderTargets.d.ts
@@ -1,0 +1,10 @@
+import { WebGLRenderTarget } from './WebGLRenderTarget';
+import { Texture } from '../textures/Texture';
+
+export class WebGLMultipleRenderTargets extends WebGLRenderTarget {
+    constructor(width: number, height: number, count: number);
+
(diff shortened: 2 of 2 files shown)


joshuaellis (dismissed):

Mugen87:
Seems the linter reports an error. Do you have an idea what is going wrong?

joshuaellis:
`WebGLMultipleRenderTargets` extends `WebGLRenderTarget` so TS expects the property `texture` to be the same type...

@Methuselah96 Do you have an ideas how we might resolve this? 🤔 I always find it odd that you can override methods but not properties.

Although @Mugen87, looking at the THREE code, wouldn't calling `WebGLMultipleRenderTargets.setTexture` fail because i'd either have to supply an array which would break:

```js
texture.image = {
	width: this.width,
	height: this.height,
	depth: this.depth
};
```

see [here](https://github.com/mrdoob/three.js/blob/d39d82999f0ac5cdd1b4eb9f4aba3f9626f32ab6/src/renderers/WebGLRenderTarget.js#L46).

Alternatively not passing an array of Textures, would break the `WebGLMultipleRenderTargets`?

Mugen87:
It seems `setTexture()` should not be used in context of `WebGLMultipleRenderTargets` since it maintains multiple textures in a special way. If you call the method with a single texture, the code will crash.

joshuaellis:
Yeah, so this afaik wouldn't be 'allowed' in typescript... So we may need to copy the class `WebGLRenderTarget` and remove certain bits as opposed to extending from... 🤔

Mugen87:
Closing this PR. I'm afraid the mentioned problems exceed my TS knowledge^^. I hope somebody else can tackle this feature.

BTW: It was considered in https://github.com/mrdoob/three.js/pull/16390#issuecomment-823994207 to not extend the class from `WebGLRenderTarget`. Maybe it's a good idea to highlight the typing issues at the `three.js` repo. Seems the class design could be cleaner if the inheritance is changed.

PR #97, opened 2021-06-23
fix: missing `.morphTargets` in MeshDepthMaterial

ycw:
### Why

There's a prop `.morphTargets` ( in MeshDepthMaterial.js https://github.com/mrdoob/three.js/blob/master/src/materials/MeshDepthMaterial.js#L32 ) which is missing in MeshDepthMaterial.d.ts 

### What

Add `.morphTargets` declaration.

### Checklist

-   [x] Ready to be merged

Thanks.


joshuaellis (approved):

PR #100, opened 2021-07-01
Update prettier from v2.2.1 to v2.3.2 and run prettier:write

bjornstar:
### Why

`package.json` says you want `prettier` ^2.2.1 and v2.3.2 is available.

`yarn.lock` has it locked to v2.2.1 causing `yarn` and `npm` to install different versions of prettier.

### What

`npx yarn upgrade prettier`
`npm run prettier:write`

If your intention is to only use v2.2.1, we can change the package.json to not request a semantic version range.

### Checklist

-   [x] Ready to be merged


Methuselah96 (changes requested):
Thanks for the contribution! Can you update the `package.json` to `prettier@2.3.2` as well? We can just lose the range since SemVer and Prettier don't really make sense in this case.

Methuselah96 (approved):
Thanks!

PR #104, opened 2021-07-13
Update IFCLoader types

agviegas:
We have added lots of new functionalities to the new version of the IFC Loader (which is [already merged](https://github.com/mrdoob/three.js/pull/22113)), and we just updated the types.


joshuaellis:
Have you ran prettier? I might add it as a husky hook to slim down this happens, I can appreciate its a bit annoying.

agviegas:
Sorry, I missed it. I'll run it! A husky hook would be nice for the future :)

agviegas:
Hum, strange. I ran prettier with Webstorm (control alt L) and `npm run prettier:write`. Is there anything else I have to do?

Methuselah96:
It's failing on lint errors now:
```
Error: /home/runner/work/three-ts-types/three-ts-types/types/three/examples/jsm/loaders/IFCLoader.d.ts:47:8
ERROR: 47:8    strict-export-declare-modifiers  'declare' keyword is redundant here. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 95:9    no-redundant-jsdoc-2             JSDoc tag '@type' is redundant in TypeScript code.
ERROR: 149:58  use-default-type-parameter       This is the default value for this type parameter, so it can be omitted.
ERROR: 167:72  use-default-type-parameter       This is the default value for this type parameter, so it can be omitted.
ERROR: 173:8   strict-export-declare-modifiers  'declare' keyword is redundant here. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 232:9   no-redundant-jsdoc-2             JSDoc tag '@type' is redundant in TypeScript code.
ERROR: 292:61  no-self-import                   Declaration file should not use a global import of itself. Use a relative import. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-self-import.md
ERROR: 311:65  no-self-import                   Declaration file should not use a global import of itself. Use a relative import. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-self-import.md
```

joshuaellis (approved):
Looks great! Is this for the current release or the upcoming new release?

agviegas:
> Looks great! Is this for the current release or the upcoming new release?

For the next release 🙂 https://github.com/mrdoob/three.js/pull/22352#issuecomment-901944515

PR #106, opened 2021-07-17
MeshPhysicalMaterial: Optional constructor parameters

donmccurdy:
Material constructor parameters argument is optional; changing in MeshPhysicalMaterial for consistency with others, eg.:

https://github.com/three-types/three-ts-types/blob/master/types/three/src/materials/MeshStandardMaterial.d.ts#L42


joshuaellis (approved):

PR #107, opened 2021-07-19
refactor: Create single utility type to encapsulate color argument

### LINKED ISSUE #29: Create single utility type to encapsulate color argument

Certain functions & properties have the type `Color | string | number` because they create a `THREE.Color` in the function and those are the arguments that `THREE.Color` can take, like the example below:

```ts
    renderPass: (
        renderer: WebGLRenderer,
        passMaterial: Material,
        renderTarget: WebGLRenderTarget,
        clearColor: Color | string | number,
        clearAlpha: Color | string | number,
    ) => void;
```

### Suggested implementation:

Therefore we want to create a utility type so instead of typing `Color | string | number` we only need to write one, this would then need to replace all instances of the previous type across the codebase.

shuta13:
### What

Fix: https://github.com/three-types/three-ts-types/issues/29

### Checklist

-   [ ] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -196,6 +196,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "shuta13",
+      "name": "Shuta Hirai",
+      "avatar_url": "https://avatars.githubusercontent.com/u/38882716?v=4",
+      "profile": "https://studio.did0es.me/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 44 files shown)


joshuaellis (approved):
This looks good to me @Methuselah96 it would be good to get your opinion too 👍🏼

Methuselah96 (approved):
Looks good to me. `CapsulatedColor` probably isn't my preferred term because "capsulated" makes me think there's a value "capsulated" inside an object instead of a union, but I can't think of any obviously better names at the moment.

joshuaellis:
`UnitedColor`? or `UColor`?

Methuselah96:
Yeah, either of those work for me. Maybe `ColorRepresentation` (although that's a bit longer)?

joshuaellis:
I prefer `ColorRepresentation` @shuta13 would you mind making that change please? 🙏🏼

shuta13:
@Methuselah96 @joshuaellis 

OK, changed. Please check out! : https://github.com/three-types/three-ts-types/pull/107/commits/7aca3fbf31c8a2759dd5a4dd0fd2437682dc0de9

Methuselah96 (approved):
🎉

joshuaellis (approved):

PR #110, opened 2021-07-31
Update Water.d.ts

puxiao:
### Why

https://github.com/mrdoob/three.js/blob/dev/examples/webgl_shaders_ocean.html

line 111:

```
water.material.uniforms[ 'sunDirection' ].value.copy( sun ).normalize();
```

<br>

line 188:

```
water.material.uniforms[ 'time' ].value += 1.0 / 60.0;
```

<br>

If using typescript...

```
Property 'uniforms' does not exist on type 'Material | Material[]'.
  Property 'uniforms' does not exist on type 'Material'.
```

<br>

<br>

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/objects/Water.js

line 223:

```
const material = new ShaderMaterial( { ...

scope.material = material;
```

<br>

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/objects/Water2.js

line 92:

```
this.material = new ShaderMaterial( { ...
```


joshuaellis:
I think you need to run `prettier:write` 😄

joshuaellis (approved):

PR #111, opened 2021-08-01
feat: add init function to MarchingCubes

servinlp:
### Why

When working with the MarchingCubes I noticed the init function was missing.

### What

Added the function and ran the test.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution!

PR #116, opened 2021-08-18
Improve node material typings

xawill:
- Fix `Node` import in `JoinNode`
- Allow the use of `UVTransformNode` as `uv`for `TextureNode`. It is not obvious from sources that this is possible, but [this three.js example](https://github.com/mrdoob/three.js/blob/master/examples/webgl_materials_nodes.html) showcases it and I tested it myself.

### DIFF
--- a/types/three/examples/jsm/nodes/inputs/TextureNode.d.ts
+++ b/types/three/examples/jsm/nodes/inputs/TextureNode.d.ts
@@ -4,12 +4,13 @@ import { InputNode } from '../core/InputNode';
 import { NodeBuilder } from '../core/NodeBuilder';
 import { Node } from '../core/Node';
 import { UVNode } from '../accessors/UVNode';
+import { UVTransformNode } from '../utils/UVTransformNode';
 
 export class TextureNode extends InputNode {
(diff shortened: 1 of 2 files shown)


joshuaellis:
Is it possible to add a test for this?

xawill:
@joshuaellis Well, for the Node import in JoinNode, it is obvious that the Node import was missing and that TS was picking the one from lib.dom.d.ts. If the `Node` class in Three.js was called differently, `JoinNode` would not compile.
Regarding the `UVTransformNode`, I don't really no how to test this. And I didn't find any existing tests for the node material. Actually, I didn't find any tests for any example. You test only the core, right ? Anyway, this is just allowing `UVTransformNode` in addition to `UVNode`. It won't break any existing code.

joshuaellis:
No we test as much as I can, typically I copy the example html from three and change it to TS 🤔 although there could be a better way. To be honest I think we're missing a lot of Node stuff anyway...

xawill:
Okey, I'll try to update the PR with some tests when I can. Don't expect it too soon, though.

joshuaellis (approved):

PR #119, opened 2021-08-28
fix: data of DataTexture.constructor is okay to be null

0b5vr:
### Why

I could not put `null` to the `data` of `DataTexture.constructor` .

In fact, its default value is `null` .

https://github.com/mrdoob/three.js/blob/e62b253081438c030d6af1ee3c3346a89124f277/src/textures/DataTexture.js#L6

### What

I've changed the type of the `data` from `BufferSource` to `BufferSource | null`

### Checklist

-   [x] Ready to be merged


types/three/src/textures/DataTexture.d.ts
```diff
@@ -17,7 +17,7 @@ export class DataTexture extends Texture {
      * @param [encoding=THREE.LinearEncoding]
      */
     constructor(
-        data: BufferSource,
+        data: BufferSource | null,
```
joshuaellis: looking at the code this should be `data?: BufferSource | null | undefined`, and then the below, `width` and `height` should be `?: number | undefined`

joshuaellis (changes requested):
Thanks for raising, I think we can make this a bit better

types/three/src/textures/DataTexture.d.ts
0b5vr: ohhh that's true it's optional, thanks. why I need to specify `undefined` here though?

0b5vr: https://github.com/mrdoob/three.js/blob/14663f58916d66d5bfe6b0d6dfbe5aa701c80408/src/renderers/WebGLRenderer.js#L2108

here is one of where `width` and `height` are referenced. I don't think they are okay to be `undefined` .

```typescript
// Ref: C:\Users\yutaka\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\node_modules\typescript\lib\lib.dom.d.ts , L17095
texSubImage2D(target: GLenum, level: GLint, xoffset: GLint, yoffset: GLint, width: GLsizei, height: GLsizei, format: GLenum, type: GLenum, pixels: ArrayBufferView | null): void;
```

0b5vr: https://github.com/three-types/three-ts-types/pull/119/commits/06ae0ed6e38f707d797a6de6b2ac85eb2e1c6c56

Changed all the parameters to be optional :pray:

joshuaellis (approved):

PR #120, opened 2021-08-29
Fix type of color parameter of Reflector

0b5vr:
### Why

Because I could not put `0x7f7f7f` into `color` parameter of `Reflector`

### What

- Change the type of `color` parameter of `Reflector` and `Refractor`
    - `Color` -> `ColorRepresentation`
    - Ref: https://github.com/mrdoob/three.js/blob/master/examples/js/objects/Reflector.js#L10
    - Ref: https://github.com/three-types/three-ts-types/blob/master/types/three/src/math/Color.d.ts#L19
- Also add a test, `objects-reflector.ts`

### Checklist

    - It's already there
-   [x] Ready to be merged


joshuaellis (approved):
Thanks for adding a test too! 💪🏼

PR #125, opened 2021-09-10
Make EventDispatcher generics

woo-cie:
This PR is ported from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55657

### Why
1. Collect type of event.target
Now, event.target is defined as any or undefined in listener args.
But, in js implementation, event.target must be EventDispatcher itself.
So, it has to be defined as this.

2. Enable to restrict method's args by type
Current declaration allows code below.
``` ts
class MyEventDispatcher extends EventDispatcher {
  someFunction(){
    this.dispatch({event: "onSomeEvent", attachment: 10});
  }
}

// somewhere
myEventDispatcher.addEventListener("onSomeEvent", e => console.log(e.attachmant))
// expected 10 but actually undefined because of typo
```
So I enable to restrict args like this.
``` ts
type Event = {type: "onSomeEvent", attachment: number} | {type: "onSomeEvent2", attachment2: string};
class MyEventDispatcher extends EventDispatcher<Event> {
  someFunction(){
    this.dispatch({event: "onSomeEvent", attachment: 10});
  }
}

// somewhere
myEventDispatcher.addEventListener("onSomeEvent", e => console.log(e.attachment))
// Auto-completion also works well.

myEventDispatcher.addEventListener("onSomeEvent2", e => console.log(e.attachment))
// Error: Property 'attachment' does not exist on ...
```

### What
This PR
- add customizable BaseEvent interface
- add EventListener type
- make EventDispatcher generics

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -214,6 +214,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "woo-cie",
+      "name": "Makoto Yamada",
+      "avatar_url": "https://avatars.githubusercontent.com/u/24642989?v=4",
+      "profile": "https://github.com/woo-cie",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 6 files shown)


Methuselah96 (approved):
This strikes the perfect balance of preserving old behavior and expanding the types to allow for better typing. Good work!

PR #130, opened 2021-09-19
Make Raycaster::Intersection, intersectObject generic

Degubi:
### Why

In our codebase we use jsdocs everywhere and anywhere we use the intersectObject(s) methods we have to do type casts or ts-ignores.
Example method from our code:

```javascript
/**
 * @param { Mesh<BufferGeometry, MeshStandardMaterial | MeshPhongMaterial>[] } meshesToRaytrace
 * @returns { T | null }
 */
getMeshAtCursor(meshesToRaytrace) {
	const intersects = raycaster.intersectObjects(meshesToRaytrace);

	// @ts-ignore We know this is a Mesh because we pass in meshes for checking in the first place
	return intersects.length > 0 ? intersects[0].object : null;
}
```

Here we pass in Mesh objects, but still get an Object3d back after the intersectObjects call, which is correct, but the type could be more concrete so that we don't need to do a ts-ignore.

### What

- Made Raycaster::Intersection generic
- Made Raycaster::intersectObject and Raycaster::intersectObjects generic
- Update files using Raycaster::Intersection as a return type

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -225,7 +225,7 @@
       ]
     },
     {
-          "login": "woo-cie",
+      "login": "woo-cie",
       "name": "Makoto Yamada",
       "avatar_url": "https://avatars.githubusercontent.com/u/24642989?v=4",
       "profile": "https://github.com/woo-cie",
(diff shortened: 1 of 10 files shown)


joshuaellis (approved):
Great change, thank you for the contribution ⭐

PR #131, opened 2021-09-23
fix: add null to Material.shadowSide's type

WCWedin:
It is currently not possible to work with `Material.shadowSide`'s default value of null, as property's type does not include `null`.

This commit changes the type of `Material.shadowSide` from `Side` to `Side | null`.

-   [x] Ready to be merged


joshuaellis (approved):

PR #133, opened 2021-09-25
Update MeshPhysicalMaterial.d.ts

donmccurdy:
Marks `sheenTint` as non-nullable, renames `attenuationColor` -> `attenuationTint`.

https://github.com/mrdoob/three.js/blob/e62b253081438c030d6af1ee3c3346a89124f277/src/materials/MeshPhysicalMaterial.js#L71-L79


joshuaellis (approved):
thanks for the contribution! ⭐

PR #138, opened 2021-10-06
r133 - include default Object3D in Intersection generic type

dbuck:
### Why
Adding a default type of Object3D to the newly generic Intersection type, and simplifying the Array<Intersection<Object3D>> types back to Intersection[] again.

Reduces churn in client code which may already reference the Intersection type.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -259,6 +259,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "dbuck",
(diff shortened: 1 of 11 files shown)


joshuaellis (approved):
This should have been caught in the PR that introduced this change, so thank you for fixing my mistake 👍🏼

dbuck:
@joshuaellis I added the yarn.lock file change which automation was complaining about

PR #141, opened 2021-10-15
fix: Adding CustomToneMapping constant to ToneMapping enum.

WCWedin:
### What

The CustomToneMapping mapping constant isn't present in the ToneMapping enum definition, making the value of the constant inaccessible. This pull request adds the value to the enum.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/constants.d.ts
+++ b/types/three/src/constants.d.ts
@@ -105,6 +105,7 @@ export const LinearToneMapping: ToneMapping;
 export const ReinhardToneMapping: ToneMapping;
 export const CineonToneMapping: ToneMapping;
 export const ACESFilmicToneMapping: ToneMapping;
+export const CustomToneMapping: ToneMapping;
 
 // Mapping modes
 export enum Mapping {}


joshuaellis:
Thanks for the contribution, do you have a link to the `three` repo that addresses this? 😄

joshuaellis:
Aren't these `ToneMapping` constants supposed to be `number`s?

WCWedin:
Regarding your first question, the constant doesn't seem to be mentioned in the three.js documentation, but it does feature in the related example: https://github.com/mrdoob/three.js/blob/master/examples/webgl_tonemapping.html
As to your second, the constants exposed by three.js are themselves numbers, yes (https://github.com/mrdoob/three.js/blob/dev/src/constants.js#L55), but they're being assigned empty enum types here, as far as I can tell, primarily as marker types for type safety, correct? I don't think I fully understand the question, though.

WCWedin:
Hi, @joshuaellis! Just a post-holiday bump in case you missed my earlier reply. Apologies and thanks!

joshuaellis (approved):

joshuaellis:
Hey sorry & thank you for the bump. Lets keep it how it is now & therefore merge what you have 👍🏼

PR #142, opened 2021-10-19
Add `createShaderMaterial` to `GPUComputationRenderer`

mattrossman:
### Why

See the JS definition here: https://github.com/mrdoob/three.js/blob/r133/examples/jsm/misc/GPUComputationRenderer.js#L302-L318

This function's usage is also mentioned in the comments at the top of the file: https://github.com/mrdoob/three.js/blob/r133/examples/jsm/misc/GPUComputationRenderer.js#L84-L85

### What

Adds the missing `createShaderMaterial` function definition to `GPUComputationRenderer`.

I didn't see an existing GPGPU test and I'm reluctant to write my own since there's a lot of moving parts and shader fluff in the [three.js examples for those](https://github.com/mrdoob/three.js/blob/d9af9952fc32c2a1281707946177135b90297aab/examples/webgl_gpgpu_water.html), seems like it'd be a lot of work to properly cover all the parts of GPUComputationRenderer. If you have a suggestion for a simple test case then I can add it though.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/OTHER_FILES.txt
+++ b/types/three/OTHER_FILES.txt
@@ -83,9 +83,7 @@ examples/jsm/math/Lut.d.ts
 examples/jsm/math/MeshSurfaceSampler.d.ts
 examples/jsm/math/OBB.d.ts
 examples/jsm/math/Octree.d.ts
-examples/jsm/math/SimplexNoise.d.ts
 examples/jsm/misc/ConvexObjectBreaker.d.ts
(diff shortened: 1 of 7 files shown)


joshuaellis:
I've added a test and removed what I believe to be a weakness in our type system regards to `RenderTarget` a generic empty object. The test highlighted this for me which is nice.

joshuaellis (approved):

PR #143, opened 2021-10-20
Add gltf loader types

robertlong:
### Why

This PR adds types to `GLTFParser` and `GLTFLoaderPlugin` which makes it easier to author new GLTFLoader plugins in typescript.

### What

This PR includes additional type definitions for the [GLTFLoader](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js) file.

I added types for properties to the `GLTFParser` such as `options` which is useful for accessing the `options.path` variable to determine the base path to use in a plugin. Or the `textureLoader` which is useful when you want to reuse the existing texture loader.

I also added the `createNodeMesh` hook to the `GLTFLoaderPlugin` which is used to create different mesh objects.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/types/three/examples/jsm/loaders/GLTFLoader.d.ts
@@ -13,6 +13,9 @@ import {
     Material,
     SkinnedMesh,
     Texture,
+    TextureLoader,
+    FileLoader,
+    ImageBitmapLoader,
 } from '../../../src/Three';
 
 import { DRACOLoader } from './DRACOLoader';
@@ -75,6 +78,19 @@ export interface GLTFReference {
 export class GLTFParser {
     json: any;
 
+    options: {
+        path: string;
+        manager: LoadingManager;
+        ktx2Loader: KTX2Loader;
+        meshoptDecoder: /* MeshoptDecoder */ any;
+        crossOrigin: string;
+        requestHeader: { [header: string]: string };
+    };
+
+    fileLoader: FileLoader;
+    textureLoader: TextureLoader | ImageBitmapLoader;
+    plugins: GLTFLoaderPlugin;
+    extensions: { [name: string]: any };
     associations: Map<Object3D | Material | Texture, GLTFReference>;
 
     getDependency: (type: string, index: number) => Promise<any>;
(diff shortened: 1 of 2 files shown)


types/three/examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -75,6 +78,19 @@ export interface GLTFReference {
 export class GLTFParser {
     json: any;
 
+    options: {
+        path: string;
+        manager: LoadingManager;
+        ktx2Loader: KTX2Loader;
+        meshoptDecoder: /* MeshoptDecoder */ any;
```
joshuaellis, line 85: could you explain why this is `any`?

types/three/examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -75,6 +78,19 @@ export interface GLTFReference {
...
+        crossOrigin: string;
+        requestHeader: { [header: string]: string };
+    };
+
+    fileLoader: FileLoader;
+    textureLoader: TextureLoader | ImageBitmapLoader;
+    plugins: GLTFLoaderPlugin;
+    extensions: { [name: string]: any };
```
joshuaellis, line 93: could you explain why this is `any`?

joshuaellis (changes requested):
Two discussion points, also if you could reference the source of these changes (a link to the file in the three repo) that would be super helpful for us ⭐

types/three/examples/jsm/loaders/GLTFLoader.d.ts
robertlong: Currently there is no MeshoptDecoder type definitions and the [setMeshoptDecoder](https://github.com/three-types/three-ts-types/blob/master/types/three/examples/jsm/loaders/GLTFLoader.d.ts#L56) method already takes an any value so I stuck with that convention.

types/three/examples/jsm/loaders/GLTFLoader.d.ts
robertlong: The extensions object currently holds [booleans for the new plugins](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L335) and [various extension objects for the older extensions](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L349). In plugins you likely want to check for the existence of a given extension. Typing all the older extensions would likely be a waste of time because they are deprecated and likely to be ported to the newer plugin api soon enough.

robertlong:
Hopefully those comments and the link in the description should help. Let me know if there's anything else.

types/three/examples/jsm/loaders/GLTFLoader.d.ts
joshuaellis: Makes sense with me!

joshuaellis (approved):

PR #151, opened 2021-11-29
feat: Add ShaderMaterial uniforms typing and autocompletion (+ ShaderPass)

Lutymane:
### Why

To increase quality and speed of development. With autocompletion you won't misspell uniform names and mistype their values.

### What

Added uniforms typings for ShaderMaterial, RawShaderMaterial and ShaderPass and also overall type improvements. 

The uniform types are automatically inferred when creating an instance from parameters.

image

Or they can be supplied in generic (useful when just declaring variables) in multiple ways:
```ts
let shaderObject = {
  vertexShader: "",
  fragmentShader: "",
  uniforms: {
    u_one: { value: 1 },
    u_two: { value: null },
  },
};

let s0: ShaderMaterial<typeof shaderObject>;
```

```ts
let r1: RawShaderMaterial<{ u_time: number; u_amp: number }>;

let r2: RawShaderMaterial<{ u_one: { value: number }; u_two: { value: Vector2 } }>;
```

The uniform types will be asserted while creating an instance.

In `ShaderMaterial.d.ts` you can also find a more strict variant for uniforms (which is commented out) for the case when you don't supply types during declaration, meaning uniforms themselves can be undefined (as there's no assertion during instance creation) or when you cast Material to ShaderMaterial which produces a correct type error but due to the fact this practice is used everywhere in legacy code I decided not to keep it and leave it more loosely making it fully backwards compatible.

Also as `null` a separate type it's recommended to type uniforms the following way:
```ts
new ShaderMaterial({
  uniforms: {
    uTexture: { value: null as Texture | null },
  },
  vertexShader: "",
  fragmentShader: "",
});
```
Because if it's simply a `null` it will be replaced by `any` to allow value modification.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -286,6 +286,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "Lutymane",
+      "name": "Lutymane",
+      "avatar_url": "https://avatars.githubusercontent.com/u/22231294?v=4",
+      "profile": "https://zeokku.com/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
   </tr>
   <tr>
     <td align="center"><a href="https://xk.io/"><img src="https://avatars.githubusercontent.com/u/1046448?v=4?s=100" width="100px;" alt=""/><br /><sub><b>Max Kaye</b></sub></a><br /><a href="https://github.com/three-types/three-ts-types/commits?author=XertroV" title="Documentation">📖</a></td>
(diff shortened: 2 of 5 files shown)


joshuaellis:
This PR has become a bit convoluted, I think if you want to fix some types these should be different PRs

Lutymane:
Oh, I'm sorry it just took so long I started to forget about this PR so it became indeed convoluted, sorry again. Tell me, please, if the part about shader materials is ok, so I'll isolate it. While I'll create a separate PR for postprocessing changes on a different branch

0b5vr:
I just have reached to the same `TUniforms` idea, while I rather want to change the definition of `Shader` which id defined in `types/three/src/renderers/shaders/ShaderLib.d.ts` .

alvinleung:
What is the status of this pr? I would love to use it on my project

PR #169, opened 2022-01-29
chore(WebXRManager): add missing methods, update getCamera signature

CodyJasonBennett:
### Why

Some changes to WebXRManager's signature weren't tracked with https://github.com/mrdoob/three.js/pull/21886 and various other changes with the introduction of layers and controllers.

### What

This PR updates the getCamera signature to return an XRCamera and introduces low-level getBaseLayer, getBinding, getFrame, updateCamera, and setReferenceSpace methods. I've vendored [Babylon's WebXR types](https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/LibDeclarations/webxr.d.ts) and updated WebXRManager and WebXRController.

### Checklist

-   [x] Ready to be merged


types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -26,6 +30,10 @@ export type XRAnimationLoopCallback = (time: number, frame?: XRFrame) => void;
 
 export type XRFrameRequestCallback = (time: number, frame: XRFrame) => void;
 
+export type XRCameraL = PerspectiveCamera & { viewport: Vector4 };
+export type XRCameraR = PerspectiveCamera & { viewport: Vector4 };
+export type XRCamera = Omit<ArrayCamera, 'cameras'> & { cameras: [XRCameraL, XRCameraR] };
```
CodyJasonBennett: Viewport is [manually instantiated when creating the ArrayCamera](https://github.com/mrdoob/three.js/blob/b5594736e654ce7c7f45e55492f6b72615b53c0d/src/renderers/webxr/WebXRManager.js#L51-L63) and is [set from XRViewport's properties](https://github.com/mrdoob/three.js/blob/b5594736e654ce7c7f45e55492f6b72615b53c0d/src/renderers/webxr/WebXRManager.js#L640) from the base layer but I used three's math classes just to be sure.

types/three/src/renderers/webxr/WebXRManager.d.ts
```diff
@@ -16,15 +26,24 @@ export class WebXRManager extends EventDispatcher {
...
     isPresenting: boolean;
 
+    /**
+     * @default true
+     */
+    cameraAutoUpdate: boolean;
+
     getController(index: number): Group;
     getControllerGrip(index: number): Group;
     getHand(index: number): Group;
```
CodyJasonBennett: These are not documented/typed, but they come from WebXRController's various space methods:
 - https://github.com/mrdoob/three.js/blob/dev/src/renderers/webxr/WebXRController.js#L16

I have left them as-is until I (or someone) documents them in three just in case anything looks wrong. There's a lot of stuff missing in this department.

types/three/src/renderers/webxr/WebXR.d.ts
0b5vr: `WebXR.d.ts` is for covering [WebXR Device API](https://developer.mozilla.org/ja/docs/Web/API/WebXR_Device_API) I think.

I believe this file should not depend on any Three.js components.
You probably have to choose another place to put these type definitions.

types/three/src/renderers/webxr/WebXRManager.d.ts
```diff
@@ -16,15 +26,24 @@ export class WebXRManager extends EventDispatcher {
...
     setFramebufferScaleFactor(value: number): void;
     setReferenceSpaceType(value: XRReferenceSpaceType): void;
     getReferenceSpace(): XRReferenceSpace | null;
+    getBaseLayer(): XRWebGLLayer | any // XRProjectionLayer
+    getBinding(): any // XRWebGLBinding
```
0b5vr: `XRWebGLBinding` is not defined even in `@types/webxr` :weary:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webxr/index.d.ts

types/three/src/renderers/webxr/WebXR.d.ts
0b5vr: @joshuaellis any idea? just moving them into `WebXRManager.d.ts` would be good imo.
I would also rename this to `WebXRArrayCamera` .

types/three/src/renderers/webxr/WebXRManager.d.ts
CodyJasonBennett: Indeed, their types are rather neglected atm. See my original comment, fixed links to Babylon's types which are more correct. Can possibly vendor those until we can get these fixed upstream.

types/three/src/renderers/webxr/WebXR.d.ts
joshuaellis: I'm happy for this to be community led, I have never use the WebXR stuff. Will leave with you, @CodyJasonBennett & @0b5vr.

0b5vr: https://github.com/three-types/three-ts-types/pull/169/commits/c2535bb6d8acef6534d2d755d00c54df273b7046 : it does look good for me 🙏

tongtwist (approved):
I needed these types updates too.
Everything sounds good to me.
Btw, with these changes, types better reflects the r140 documentation of WebXRManager.getCamera()

CodyJasonBennett:
I went ahead and fixed the webxr types and completed three's controller types.

Notably, this drops `WebGLRenderingContext#makeXRCompatible` and swaps `XRGamepad` for the native `Gamepad`.

joshuaellis (approved):

joshuaellis:
@0b5vr if you approve I will merge this.

joshuaellis:
#218 merged added `@types/webxr` i'm wondering if this is still necessary? If so, the merge conflicts will need to be resolved.

CodyJasonBennett:
The changes in my PR *were* more correct than `@types/webxr` and still Babylon. I'd need to find time to review.

0b5vr:
> @0b5vr if you approve I will merge this.
 
I didn't noticed to this! Let me see this real quick.

types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -1,56 +1,111 @@
...
+ * Available session modes
+ */
 export type XRSessionMode = 'inline' | 'immersive-vr' | 'immersive-ar';
 
-export type XRReferenceSpaceType = 'viewer' | 'local' | 'local-floor' | 'bounded-floor' | 'unbounded';
+/**
+ * Reference space types
+ */
+export type XRReferenceSpaceType = 'viewer' | 'local' | 'local-floor' | 'unbounded';
+export type XRBoundedReferenceSpaceType = 'bounded-floor';
```
0b5vr: calling `requestReferenceSpace` with `bounded-floor` returns `XRBoundedReferenceSpace` instead of `XRReferenceSpace` .
Strictly, this type does not exist in WebXR but the naming and typing feels reasonable enough.

See: https://developer.mozilla.org/en/docs/Web/API/XRSession/requestReferenceSpace

types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -1,56 +1,111 @@
...
     | 'select'
     | 'selectstart'
     | 'selectend'
     | 'squeeze'
     | 'squeezestart'
     | 'squeezeend'
-    | 'inputsourceschange';
+    | 'reset'
+    | 'eyetrackingstart'
+    | 'eyetrackingend';
```
0b5vr: [QUESTION] What is these event types? I could not see any docs about them.

types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -1,56 +1,111 @@
...
 
 export interface XR extends EventTarget {
     requestSession(mode: XRSessionMode, options?: XRSessionInit): Promise<XRSession>;
     isSessionSupported(mode: XRSessionMode): Promise<boolean>;
 }
 
+export interface Constructor<T = object> {
+    new (...args: any[]): T;
+    prototype: T;
+}
```
0b5vr: [QUESTION] What is the purpose of this change? Is this a valid change to have in this package?

types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -1,56 +1,111 @@
...
 export interface Window {
-    XRSession?: Constructor<XRSession> | undefined;
-    XR?: Constructor<XR> | undefined;
+    XRSession?: Constructor<XRSession>;
+    XR?: Constructor<XR>;
 }
 
 export interface Navigator {
-    xr?: XR | undefined;
+    xr?: XR;
```
0b5vr: [SHOULD] `navigator.xr` still does not exist in browsers like Safari right? I prefer having `| undefined` here.

types/three/src/renderers/webxr/WebXR.d.ts
```diff
@@ -60,136 +115,298 @@ export interface XRViewport {
...
-    createAnchor(pose: XRRigidTransform, space: EventTarget): Promise<XRAnchor>;
-    // Planes
-    worldInformation: {
-        detectedPlanes?: XRPlaneSet | undefined;
-    };
+    trackedAnchors?: XRAnchorSet;
+    createAnchor?(pose: XRRigidTransform, space: XRSpace): Promise<XRAnchor>;
+
+    // World geometries. DEPRECATED
+    worldInformation?: XRWorldInformation;
```
0b5vr: oooh it's now a deprecated interface (took place by `xrFrame.detectedPlanes`). I should revise my code later...

types/three/src/renderers/webxr/WebXR.d.ts
CodyJasonBennett: These are leftovers from Babylon's native runtime.

types/three/src/renderers/webxr/WebXR.d.ts
CodyJasonBennett: Undefined is redundant with the ? modifier in TS.

CodyJasonBennett:
The updated `@types/webxr` look good to me. I'll revert related changes in my PR, but #218 didn't touch the three interfaces I originally corrected here.

Am I correct in assuming that these types are ambient? I've removed imports to the WebXR definitions accordingly.

types/three/src/renderers/webxr/WebXR.d.ts
CodyJasonBennett: See the implementation, this is not true.

types/three/src/renderers/webxr/WebXRController.d.ts
```diff
@@ -1,13 +1,42 @@
 import { Group } from '../../objects/Group';
+import { Vector3 } from '../../math/Vector3';
 
 export type XRControllerEventType = XRSessionEventType | XRInputSourceEventType | 'disconnected' | 'connected';
 
+export interface XRHandJoints {
```
joshuaellis: Maybe this should be

```ts
export type XRHandJoints = Record<XRHandJoint, number>
```

_maybe_?

joshuaellis (commented):
cc @CodyJasonBennett see above comment

types/three/src/renderers/webxr/WebXRController.d.ts
CodyJasonBennett: Is there a way to make this partial later? Controllers will only cache used joints, so I previously used `Partial` with a typed joint object.

Basically, would like `XRHandJoints` to be the above, yet `Record<XRHandJoint, number | undefined>` later. Notably, the structure of the joints object would change at runtime. It is not known which joints are tracked ahead of time.

CodyJasonBennett: Ah, seems I can still use Partial with Record.

joshuaellis (approved):
Thanks for all your work @CodyJasonBennett & thanks for the reviewing @0b5vr.

PR #171, opened 2022-02-12
Remove `enableKeys` from OrbitControls

ghost:
### Why

This is no longer being used, Ctrl+F it in https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/OrbitControls.js.

For all instances, see https://github.com/mrdoob/three.js/search?q=enableKeys.
It only seems to be used in the experimental `CameraControls`.

### What

I've removed it.

### Checklist

-   [X] Ready to be merged


joshuaellis (approved):

PR #173, opened 2022-02-15
Fix documentation for a `recursive` parameter

ghost:
This has changed in https://github.com/mrdoob/three.js/pull/22460.

It'd be great if documentation could somehow automatically be pulled from https://github.com/mrdoob/three.js. There is probably a lot of outdated documentation.


joshuaellis (approved):
I don't know of a way to pull the jsdoc comments unfortunately. If you do have an idea, please make a PR!

PR #177, opened 2022-02-21
Chore/add missing encoding constants

schwyzl:
### What

Added missing texture encodings to constants


joshuaellis:
Can you point us to the source for this change please?

0b5vr:
I'm pretty sure these are rather removed in r136:

image

https://github.com/mrdoob/three.js/releases/tag/r136

0b5vr:
you might want to use a previous version of `@types/three` which matches to the `three` version you are using @schwyzl

schwyzl:
durp

0b5vr:
it's difficult to catch up with three.js breaking changes tbh, it's alright

PR #179, opened 2022-02-22
Remove VRMLoader

0b5vr:
### Why

To catch up with r138 changes

### What

Remove VRMLoader.

See: https://github.com/mrdoob/three.js/pull/23414

(clarification: it was an implementation of VRM which was not maintained. as mrdoob mentions, people should use @pixiv/three-vrm instead.)

### Question

What was the purpose of `OTHER_FILES.txt` again?
I kinda want to document this in a separated PR since I forgot 😅 

### Checklist

    - it's a r138 change so it should be directed to `dev`
-   [x] Ready to be merged


joshuaellis (approved):
It's basically to say that these files are included in the type system but aren't accessed by the source, i'm not sure more than that unfortunately.

PR #180, opened 2022-02-22
feature (Texture): Add Source class

0b5vr:
### Why

To catch up with r138 changes

### What

Added `Source` class that is going to be introduced in r138.

I also added a test code `textures-source.ts` .

See: https://github.com/mrdoob/three.js/pull/22846
See: https://github.com/mrdoob/three.js/pull/23419

### Points need review

- [x] ~~I have very low confidence about the signature of `Source.data` and `Texture.image`~~
    - ~~@Mugen87 would you mind reviewing this?~~
      ~~I don't fully understand how `Source` works with video textures, compressed textures, ...~~
    - I decided to make it `any` for now.
- [x] ~~Is it reasonable to define a type `SourceData` ?~~

### Checklist

    - It's a r138 change
-   [ ] Ready to be merged

### DIFF
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -225,4 +225,5 @@ export * from './textures/CubeTexture';
 export * from './textures/CanvasTexture';
 export * from './textures/DepthTexture';
 export * from './textures/FramebufferTexture';
+export * from './textures/Source';
 export * from './textures/Texture';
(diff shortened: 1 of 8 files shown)


0b5vr:
~~considering having `<TSource = any>` generics on Texture and Source......~~
I should do this in a separated PR.

Let me make a PR using `any` for this one.

joshuaellis (approved):
Great work adding a test!

PR #184, opened 2022-03-01
change: Added Vector3.setFromEuler(); removed Euler.toVector3()

0b5vr:
### Why

To catch up with r138

### What

See: https://github.com/mrdoob/three.js/pull/23494

- Add `Vector3.setFromEuler`
- Make `Euler.toVector3` deprecated

### Points need review

- [ ] `Euler.toVector3` actually does not fallback to the new `Vector3.setFromEuler` ([code](https://github.com/mrdoob/three.js/pull/23494/files#diff-5b25cb6d07a2d314fa35eb2138d99f2484837dcc079fd06e0cb108e8aa70618aR405-R409)). Is this okay to mark this be deprecated or should I remove the signature instead?

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/math/Euler.d.ts
+++ b/types/three/src/math/Euler.d.ts
@@ -38,7 +38,12 @@ export class Euler {
     equals(euler: Euler): boolean;
     fromArray(xyzo: any[]): Euler;
     toArray(array?: number[], offset?: number): number[];
+
+    /**
+     * @deprecated .toVector3() has been removed. Use {@link Vector3.setFromEuler Vector3.setFromEuler()} instead
+     */
     toVector3(optionalResult?: Vector3): Vector3;
+
     _onChange(callback: () => void): this;
 
     static RotationOrders: string[];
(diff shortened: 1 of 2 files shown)


types/three/src/math/Euler.d.ts
```diff
@@ -38,7 +38,12 @@ export class Euler {
     equals(euler: Euler): boolean;
     fromArray(xyzo: any[]): Euler;
     toArray(array?: number[], offset?: number): number[];
+
+    /**
+     * @deprecated .toVector3() has been removed. Use {@link Vector3.setFromEuler Vector3.setFromEuler()} instead
+     */
     toVector3(optionalResult?: Vector3): Vector3;
```
joshuaellis, line 45: If they've removed it from the source, we should remove it here.

joshuaellis (changes requested):

joshuaellis:
closed in favour of b1f8b5f

PR #197, opened 2022-03-11
[postprocessing] Some SSAARenderPass arguments are optional

willstott101:
### Why

[SSAARenderPass implementation is robust to undefined for the last two arguments](https://github.com/mrdoob/three.js/blob/master/examples/jsm/postprocessing/SSAARenderPass.js#L36)

### Checklist

-   [ ] Added myself to contributors table - not really bothered tbh, github keeps track anyway?
-   [x] Ready to be merged


joshuaellis (approved):

PR #198, opened 2022-03-16
WebGLUtils isn't exported and WebGLUtils.convert() has wrong argument type

DavidPeicho:
### Why

`WebGLUtils.d.ts` had several issues described in the [What](#what) section.

### What

1. `WebGLUtils` is exported by Three.js, but not by the typings. I personally use the `convert` method as well.
2. `WebGLUtils.convert()` accepts more than just compressed format.

### Checklist

-   [X] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -305,6 +305,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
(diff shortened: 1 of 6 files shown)


joshuaellis (approved):

PR #211, opened 2022-05-01
feat: Generate constructor interfaces

capnmidnight:
### Why

In TypeScript, the parameters to a constructor can be retrieved with a utility type called `ConstructorParameters<Type>`: https://www.typescriptlang.org/docs/handbook/utility-types.html#constructorparameterstype

The Type parameter to ConstructorParameters expects a Constructor interface type, which is an interface with a `new` member and a `prototype` member.

By defining these *Constructor interfaces, Three.js class types can now be used with ConstructorParameters.

### What

I wrote a C# program to scan all the .d.ts files for class exports, extract an interface definition for the constructors, and insert the constructor interfaces after each class definition. My program correctly handles generic classes, multiple class definitions per file, and multiple constructor overloads. You can see that program here: https://github.com/capnmidnight/Juniper/blob/master/etc/GenerateConstructorTypes/Program.cs

### Checklist

-   [ ] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -314,6 +314,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "capnmidnight",
+      "name": "Sean T. McBeth",
+      "avatar_url": "https://avatars.githubusercontent.com/u/298046?v=4",
+      "profile": "http://www.seanmcbeth.com/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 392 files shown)


joshuaellis:
I'm not sure I understand why we should do this? Also

/cc @Methuselah96 for your opinion

capnmidnight:
Well, it lets you create functions and other classes that match the signature of the target class. Maybe one would do something like:

```typescript
const cache = new WeakMap<String, THREE.SphereGeometry>();

function createSphere(...params: ConstructorParameters<THREE.SphereGeometryConstructor>): THREE.SphereGeometry {
    const key = JSON.stringify(params);
    if (cache.has(key)) {
        return cache.get(key);
    }

    const value = new THREE.SphereGeometry(...params);
    cache.set(key, value);
    return value;
}

const a = createSphere(1, 10, 5);
const b = createSphere(1, 10, 5);

assertTrue(a === b);
```

You could also do something like:
```typescript
function vector(...params: ConstructorParameters<THREE.Vector3Constructor>): THREE.Vector3;
function vector(...params: ConstructorParameters<THREE.Vector4Constructor>): THREE.Vector4;
function vector(...params: ConstructorParameters<THREE.Vector3Constructor> | ConstructorParameters<THREE.Vector4Constructor>): THREE.Vector3 | THREE.Vector4 {
    if (params.length === 3) {
        return new THREE.Vector3(...params);
    }
    else {
        return new THREE.Vector4(...params);
    }
}

const v3 = vector(1, 2, 3);
const v4 = vector(1, 2, 3, 4);

assertTrue(v3 instanceof THREE.Vector3);
assertTrue(v4 instanceof THREE.Vector4);
```

Methuselah96:
Could you do:
```ts
ConstructorParameters<typeof THREE.SphereGeometry>
```

Instead of:
```ts
ConstructorParameters<THREE.SphereGeometryConstructor>
```

capnmidnight:
Ah, I had forgotten about `typeof X`.

The constructor interface does have the additional benefit of allowing one to pass constructors as parameters and still be able to call them with `new` and construct objects with them. Maybe that isn't that important of an issue, as there are few classes in Three.js that have the same constructor signature, and of the ones that due, they are very simple.

Methuselah96:
Yeah, I don't think there's a common enough use-case to justify adding this amount of duplication at this point.

PR #213, opened 2022-05-02
Fixed SavePass constructor

maccesch:
Parameter `renderTarget` is optional in the constructor of `SavePass`.

### Checklist

  - feels a bit prentenious to add myself there because I added literally just one single character. But it's in the checklist, so I did 😄 
-   [x] Ready to be merged


joshuaellis (approved):

joshuaellis:
You need to add the test to the `tsconfig` inside `three`

joshuaellis:
Thanks! I've just added this to our r140 release, so it should be in there when thats merged.

Issue #215, opened 2022-05-11
The 140 types don't compile, while 139 does - GLTFExporter.


elalish:
-   `three` version: r140.2
-   `node` version: v16.13.2
-   `npm` version: 8.1.2

### Problem description:

I'm upgrading [model-viewer](https://github.com/google/model-viewer/pull/3411) from three.js r139 to r140.2; this is working just fine if I leave @types/three at 0.139.0. However, if I also update the types to 0.140.0, I get the following build error:
```
node_modules/@types/three/examples/jsm/exporters/GLTFExporter.d.ts:94:5 - error TS7010: 'setPlugins', which lacks return-type annotation, implicitly has an 'any' return type.

94     setPlugins(plugins: GLTFExporterPlugin[]);
```

### Suggested solution:

It's clearly a regression, and oddly it's a regression in the types themselves rather than due to the three.js upgrade, since the 139 types are working for both old and new three.js. Hopefully the errant PR shouldn't be too hard to find.

PR #220, opened 2022-05-19
Fix for issue #215: implicit any return type for GLTFExporter's setPlugin method.

capnmidnight:
### Why

As reported in issue #215, the latest version of GLTFExporter.d.ts has a setPlugin method defined without an annotation for the return type. This causes projects with "no-implicit-any" set in tsconfig to fail.

### What

The code in Three.js returns nothing, so I set the return type of the setPlugin method to `void`.

### Checklist

-   [ ] Ready to be merged


joshuaellis:
This was resolved in #214

PR #218, opened 2022-05-13
fix: WebXR examples

capnmidnight:
### Why

There are a number of incorrect type definitions in the WebXR examples code. In particular:

 - `ARButton`s optional `sessionInit` constructor parameter is specified as `any`
 - `XRControllerModelFactory` is missing an `envMap` field
 - `OculusHandModel` and `OculusHandPointerModel` have a `motionController` field that is specified as `Object3D`, when a more specific type is avaiable
 - `XREstimatedLight`'s `lightProbe` field is specified as `unknown`
 - `XRHandModelFactory`'s `createHandModel` method's optional `profile` parameter includes `"oculus"` as a valid value, when that value has not been available for over a year.

### What

 - Change the `ARButton`s `sessionInit` constructor parameter to an extension of the XRSessionInit that provides for the option of specifying a `domOverlay` setting
 - Add an `envMap: Texture` field to `XRControllerModelFactory`
 - Change the `motionController` field in `OculusHandModel` and `OculusHandPointerModel` to be `XRHandMeshModel`
 - Change the `lightProbe` field in `XREstimatedLight`'s to be Three.js' `LightProbe` type.
 - Change the `profile` parameter in `XRHandModelFactory`'s `createHandModel` method's to accept `"mesh"` in place of `"oculus"`

### Checklist

-   [x] Checked the target branch (current goes `master`, next goes `dev`)
-   [x] Added myself to contributors table
-   [ ] Ready to be merged

PR #221, opened 2022-05-20
Update webXR hand profile names

Neptilo:
### Why

The hand tracking profile name "mesh" is not considered valid because the type definitions still use the old name "oculus".
The hand profile names were changed in THREE.js r129 ([commit 571347f](https://github.com/mrdoob/three.js/commit/571347f504dec332f6905bb3d8ef43f1b405d45c)). Since this repo is supposed to correspond to revision r140, this was probably an omission and should be updated.

### What

Updated the profile name to reflect the new one

### Checklist

-   [x] Ready to be merged


joshuaellis:
this was resolved in #218, thank you for the contribution however.

PR #223, opened 2022-05-28
fix: correctly import webxr types

Methuselah96:
Resolves https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60551#issuecomment-1139372590.

### DIFF
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -207,7 +207,6 @@ export * from './renderers/webgl/WebGLShadowMap';
 export * from './renderers/webgl/WebGLState';
 export * from './renderers/webgl/WebGLTextures';
 export * from './renderers/webgl/WebGLUniforms';
-export * from './renderers/webxr/WebXR';
 export * from './renderers/webxr/WebXRController';
 export * from './renderers/webxr/WebXRManager';
 export { WebGLUtils } from './renderers/webgl/WebGLUtils.js';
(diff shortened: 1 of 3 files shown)


CodyJasonBennett:
Would this have to be mirrored in other classes that make references to WebXR types like [WebXRController](https://github.com/three-types/three-ts-types/blob/master/types/three/src/renderers/webxr/WebXRController.d.ts)?

What would this look like for user-facing code?

Methuselah96:
No, you only have to reference it in one file and then it's globally available. Users won't have to do anything different as far as I'm aware.

CodyJasonBennett:
Seems this was quite the breaking change since projects using types <141 were importing from three, and in order to support ambient types with 141 they'd require a major bump to hike the peer dep, or also include `@types/webxr` likewise.

Any chance we can revert or export `@types/webxr` in a patch release?

joshuaellis:
Sorry, I'm not sure what the problem you're trying to say is.

It sounds like to me people are installing the latest version of these types with the non equivalent version of three e.g r139, if this is the case this is an incorrect usage of the library.

Let me know if I've misunderstood?

CodyJasonBennett:
`@types/three` no longer exports WebXR types since r141. This is a breaking change.

joshuaellis:
Right, yeah if you can open a new issue for this, we can track it better there.

PR #224, opened 2022-05-29
fix: return type for GLTFLoader's parseAsync method

capnmidnight:
### Why

`GLTFLoader::parseAsync()` indicates a `Promise<void>` return type, which is incorrect

### What

Change `GLTFLoader::parseAsync()` to return `Promise<GLTF>`.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/types/three/examples/jsm/loaders/GLTFLoader.d.ts
@@ -65,7 +65,7 @@ export class GLTFLoader extends Loader {
         onError?: (event: ErrorEvent) => void,
     ): void;
 
-    parseAsync(data: ArrayBuffer | string, path: string): Promise<void>;
+    parseAsync(data: ArrayBuffer | string, path: string): Promise<GLTF>;
 }
 
 export type GLTFReferenceType = 'materials' | 'nodes' | 'textures' | 'meshes';


types/three/examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -65,7 +65,7 @@ export class GLTFLoader extends Loader {
         onError?: (event: ErrorEvent) => void,
     ): void;
```
joshuaellis, line 66: Why doesn't `parse` return a GLTF?

joshuaellis (changes requested):
More of a question than a change.

types/three/examples/jsm/loaders/GLTFLoader.d.ts
capnmidnight: It's not implemented that way https://github.com/mrdoob/three.js/blob/00d3363e2d8a6d43dfdb75e50e3ff553b3c2f932/examples/jsm/loaders/GLTFLoader.js#L278

The semantics of `parse` vs `parseAsync` in Three.js loaders is not "synchronous versus asynchronous", but "asynchronous using callbacks versus asynchronous using Promises".

joshuaellis: okay sure, but it `parseAsync` doesn't return a GLTF? it just returns a promise that resolves when `onLoad` is called in `parse` but no where does it return a `GLTF`.

capnmidnight: The promise parseAsync returns resolves to a GLTF object

joshuaellis (approved):

Issue #124, opened 2021-09-03
Add concrete types to the `image` and `mipmaps` properties of Texture


mewore:
### Describe the feature you'd like:

https://github.com/three-types/three-ts-types/blob/6a5f78138c10ca144f15f4e3ea3b989c60f307d1/types/three/src/textures/Texture.d.ts#L52
https://github.com/three-types/three-ts-types/blob/6a5f78138c10ca144f15f4e3ea3b989c60f307d1/types/three/src/textures/Texture.d.ts#L57
If it's certain that these are the only types they can have, then they can be added to the type declaration itself.

However, note that this is a **breaking change** for the transpilation of TS code which directly uses any sub-properties of these properties (except `.image.width` and `.image.height`, which are always present) or casts one of them to an unrelated type.

### Suggested implementation:

```ts
    /**
     * @default THREE.Texture.DEFAULT_IMAGE
     */
    image: HTMLImageElement | ImageData | {width: number, height: number};

    /**
     * The array of the texture mipmaps ({@link ImageData} for 2D textures and {@link CubeTexture} for cube textures).
     * 
     * @default []
     */
    mipmaps: ImageData[] | CubeTexture[];
```
This is based on my understanding that if it's a 2D texture, then *all* of its mipmaps are of type `ImageData`, and if it's a cube texture, then *all* of its mipmaps are of type `CubeTexture`. If I'm wrong, then it should be `mipmaps: (ImageData | CubeTexture)[]` instead.

capnmidnight:
The image field for a Texture can actually be a lot of different things. Just the raw types that WebGL can accept as a texture source alone are:

 - HTMImageElement
 - HTMLVideoElement
 - HTMLCanvasElement
 - OffscreenCanvas
 - ImageData
 - ImageBitmap

There is a type in `lib.d.ts` called `TexImageSource` that unions all of these types together.

But Three.js also has a few hacks in place where they use the image field to track the dimensions of the texture without actually storing a `TexImageSource`. As you've noted in your suggestion, a width/height object is one case (`CompressedTexture`, `DepthTexture`, and `FramebufferTexture`). But also, `CubeTexture` allows for an array of images, `DataTexture` is similar to `CompressedTexture` with an additional `data` field, `Data3DTexture` and `DataArrayTexture` are similar to `DataTexture` with an additional `depth` field,

[...]

PR #227, opened 2022-06-14
fix issue #124: Texture and MipMap types

capnmidnight:
### Why

Issue #124 discusses a desire to have concrete types for the `image` and `mipmaps` fields of `THREE.Texture`.

This is made difficult by the fact that Three.js appears to occasionally use the `images` field not as an actual source of image data, but as a metadata field to store width/height information about composite texture types like `THREE.CubeTexture`.

For the `mipmaps` field, all of the loaders in the examples code end up returning an `ImageData` structure, but in some cases like `THREE.VideoTexture`, the `mipmaps` field could contain an `HTMLVideoElement`. In general, Three.js does some checking on the type of the `THREE.Texture` before choosing whether to use the `image` field or one of the `mipmaps` array elements for a call to various texture loading functions in WebGL, some of which are capable of loading any `TexImageSource`. Indeed, it appears it's kept as such to also support manually generating mipmaps, e.g. by resizing images via rendering to a Canvas.

### What

For all of the Loader types in the examples folder, I've updated their returned mipmaps fields to be `ImageData[]` instead of `object[]`.

For the base `THREE.Texture` type, I started by defining a number of utility types that describe all the extant ways that the `image` field appears to be used in Three.js. I then used these to create optional type parameters to `THREE.Texture`. Subclasses of `THREE.Texture` can then specify values for those type parameters, e.g. `THREE.VideoTexture` specifies that the image type is `HTMLVideoElement`.

One curious thing is that there actually aren't a lot of restrictions on the types of images you put in `THREE.Texture`s. For example, `THREE.VideoTexture` *could* contain an `HTMLCanvasElement`. All `THREE.VideoTexture` does is run an animation loop to set `needsUpdate` without the user having to do so in their own code. You could conceivably use `THREE.VideoTexture` with a canvas object that you intend to draw to continuously, saving you from having to write an animation loop to update the texture. Similarly, `THREE.CanvasTexture` doesn't really care if you give it a canvas or not, all it does is set `needsUpdate` to true. 

But in those cases, I opted to be explicit in the typing of the textures. If someone were to want to do some shenanigans like that, they can always just cast to `any`.

Finally, I also updated the type signatures for `THREE.WebGLRenderTarget` and its subclasses, so that their `texture` fields could also be typed more explicitly.

I'm submitting this as a draft PR because I'm not really sure of the ergonomics of this. The tests pass, I even managed to write a few that prove out the general concept in type checking. But I'm not really sure how useful the generic `THREE.Texture` type with the default type parameters will be when dealing with textures loaded from model loaders. Well, at the very least, it gives you more guidance on how to go about type filtering on those values, but I'm curious to see what others think.

### Checklist

-   [ ] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -349,10 +349,8 @@
       "profile": "https://github.com/atulrnt",
       "contributions": [
         "code",
-      ],
-        "code"
       ]
-    },
+    }
   ],
   "skipCi": true,
   "contributorsPerLine": 7
--- a/types/three/examples/jsm/loaders/DDSLoader.d.ts
+++ b/types/three/examples/jsm/loaders/DDSLoader.d.ts
@@ -1,7 +1,7 @@
 import { LoadingManager, CompressedTextureLoader, PixelFormat, CompressedPixelFormat } from '../../../src/Three';
 
 export interface DDS {
-    mipmaps: object[];
+    mipmaps: ImageData[];
     width: number;
     height: number;
     format: PixelFormat | CompressedPixelFormat;
--- a/types/three/examples/jsm/loaders/KTXLoader.d.ts
+++ b/types/three/examples/jsm/loaders/KTXLoader.d.ts
@@ -1,7 +1,7 @@
 import { LoadingManager, CompressedTextureLoader, PixelFormat, CompressedPixelFormat } from '../../../src/Three';
 
 export interface KTX {
-    mipmaps: object[];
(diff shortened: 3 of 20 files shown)


Methuselah96:
Superseded by https://github.com/three-types/three-ts-types/pull/1025 and https://github.com/three-types/three-ts-types/pull/1869, although your explanation was quite helpful to read, and we may go more in this direction if it proves to be helpful.

PR #228, opened 2022-06-15
Orbital Controls modes of input to accomodate null values

Heaust-ops:
So say I want left click to control nothing.
Right now I have to do this,
```
    controls.mouseButtons = {
      LEFT: null as unknown as THREE.MOUSE,
      MIDDLE: THREE.MOUSE.ROTATE,
      RIGHT: THREE.MOUSE.LEFT,
    };
```
This PR tells typescript to expect null values when the dev wants a specific mode of input not to control anything.
'null' is probably better than making it optional / undefined because it's explicit and fields can unknowingly be left undefined.

### DIFF
--- a/types/three/examples/jsm/controls/OrbitControls.d.ts
+++ b/types/three/examples/jsm/controls/OrbitControls.d.ts
@@ -42,9 +42,9 @@ export class OrbitControls {
     autoRotate: boolean;
     autoRotateSpeed: number;
 
-    keys: { LEFT: string; UP: string; RIGHT: string; BOTTOM: string };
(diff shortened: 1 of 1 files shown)


joshuaellis:
Can you point to the THREE documentation // src code where this type change should be allowed?

Heaust-ops:
> 

Yeah, orbital controls handles these options with switch case,
like in line 868 in [OrbitControls.js](https://github.com/mrdoob/three.js/blob/master/examples/jsm/controls/OrbitControls.js),
a default case is handled, that means anything other than the expected value, like null, is handled,
and in such a case (unexpected value, like null), the orbital controls won't do anything, or sit idle, line 947.
which is exactly what we want, for them to sit idle on null.

joshuaellis:
closed in favor of #236

PR #240, opened 2022-07-30
feat: node material

### LINKED ISSUE #137: Add Nodes

https://github.com/three-types/three-ts-types/issues/129#issuecomment-932963150

TLDR; Nodes is missing in `three/examples/jsm/renderers/nodes` and has partial support in `three/examples/jsm/nodes`. Would really appreciate some help getting these in order 🙏🏼

miko3k:
### Why

I like Typescript and need the nodes material. It seems to be the only reasonable way to manage shaders. 

Issue #137 

### What

Typings for the Node material.

I also increased minimum Typescript version to 4.0 because of variadic tuple types.

Type Inference works neatly for the ShaderNode, see [an example](https://github.com/three-types/three-ts-types/pull/240/files#diff-1ce6102a82e058ac59ac2c9fbb24981c00b84495b79eba6403d9e97335e68499). It's a bit hacky, but I find it sweet. Can be changed if needed.

This PR is against master as the feature is already present in r143.

I am excited about getting some feedback to this.

A few Typescript types are defined by me in this PR:

- types defined in `nodes/core/constants.d.ts` 
- `Swizzable` and other stuff from `ShaderNode.d.ts`
- other utility types should be rather noncontroversial
  - various "scopes", they usually end with `Scope`, other examples are `MathNodeMethod`, `OperatorNodeOp`
  - search for `export type` in the source code to find out :)

### Checklist

-   [ ] Ready to be merged


types/three/tsconfig.json
```diff
@@ -71,6 +71,8 @@
         "test/utils/utils-webgl-portal.ts",
         "test/utils/utils-webgl-marchingcubes.ts",
         "test/webxr/webxr-ar-lighting.ts",
-        "test/webxr/webxr-vr-handinput-cubes.ts"
+        "test/webxr/webxr-vr-handinput-cubes.ts",
+        "test/nodes/shadernode-code-samples.ts",
```
joshuaellis: These should be in alphabetical order please!

joshuaellis:
Wow, thanks for doing so much work! I'm sure you appreciate how large this PR is and i'll need to investigate it thoroughly to get it over the line.

I think to start, i'd like to know why you've had to update the minimum version of TS to 4.0? 😄 

As a side note, typically unused files in `examples` have to be listed in `OTHER_FILES.txt` however, this is considered "bad practice" by DT and we should avoid it, to do so adding `tests` is the solution, I can see you've added some already & it'd be great to flesh more so we don't have to add any to the file, e.g. looking at the CI, `CameraNode` is not used but i'm sure we could sort this out.

miko3k:
> Wow, thanks for doing so much work! I'm sure you appreciate how large this PR is and i'll need to investigate it thoroughly to get it over the line.

Thanks, would be very nice to get feedback about this.

> I think to start, i'd like to know why you've had to update the minimum version of TS to 4.0? 😄

Bacause Variadic Tuple Types to get reasonable type inference for the ShaderMaterial. Actually take a look straight here

https://github.com/three-types/three-ts-types/pull/240/files#diff-fd87ebe379f9463f9373481514ae11c90bad2329b64743bca9b1e76270f47fd7

That would be attempt to infer types of `ShaderNode` and friends, including overloaded constructors for the `MathNode`. It's questionable if you guys wanna have something like that in tree, however as I said, I find it sweet how syntax completion work so I decided to give it a try.

The rest is rather boring. That's also the reason why I didn't cover it by tests.

> As a side note, typically unused files in `examples` have to be listed in `OTHER_FILES.txt` however, this is considered "bad practice" by DT and we should avoid it, to do so adding `tests` is the solution, I can see you've added some already & it'd be great to flesh more so we don't have to add any to the file, e.g. looking at the CI, `CameraNode` is not used but i'm sure we could sort this out.

Yes pretty much just the dreaded `ShaderNode` is tested for now.

Edit: of course I can and will gladly add more tests, I am open to suggestions how to do it properly. I could at least copy official examples of nodes here plus some random chunks of code in three.js three as I did for the`ShaderNode`. An update of the PR will follow in some hours.

types/three/tsconfig.json
miko3k: Rearranged.

Methuselah96:
> I think to start, i'd like to know why you've had to update the minimum version of TS to 4.0? 😄

That should be fine anyway, the [current support window](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window) on DefinitelyTyped is 4.0 and greater.

miko3k:
I pushed an update, `yarn run test-all` is clean now and I also do not import DOM `Node` anywhere (ups). 

Tests were extended by a few source files from three itself.

I unchecked that I am ready to merge as this obviously will need more work.

I also updated the initial description regarding the types I've taken liberty to define myself.

@joshuaellis One more note regarding your initial comment. Almost everything is used indirectly through [Nodes.d.ts](https://github.com/three-types/three-ts-types/pull/240/files#diff-7a621a9ea70ff9e6a0fc393fa3e93fed42ebed0f62b26180e68602eee24eefe3), although there are some omissions like the renderer related stuff.

types/three/examples/jsm/nodes/core/Node.d.ts
```diff
@@ -0,0 +1,29 @@
+import { NodeTypeOption, NodeUpdateTypeOption } from './constants';
+import NodeBuilder from './NodeBuilder';
+import NodeFrame from './NodeFrame';
+
+export default class Node {
+    uuid: string;
+    type: string;
+    isNode: true;
+    nodeType: NodeTypeOption;
```
joshuaellis: This can be `null`:

https://github.com/mrdoob/three.js/blob/03f35a96ae36507a2d234f6f0726dd64393ee49d/examples/jsm/nodes/core/Node.js#L9

types/three/examples/jsm/nodes/core/Node.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    id: number;
+
+    constructor(nodeType?: NodeTypeOption);
+
+    getChildren(): Node[];
+    getHash(builder: NodeBuilder): string;
+    getUpdateType(builder: NodeBuilder): NodeUpdateTypeOption;
+    getNodeType(builder: NodeBuilder, output?: string | null): NodeTypeOption | null;
+    getConstructHash(builder: NodeBuilder): string;
+    getReference(builder: NodeBuilder): Node;
```
joshuaellis, line 20: This could be `void`

types/three/examples/jsm/nodes/accessors/BufferNode.d.ts
```diff
@@ -0,0 +1,11 @@
+import { NodeTypeOption } from '../core/constants';
+import UniformNode from '../core/UniformNode';
+
+export default class BufferNode extends UniformNode {
+    isBufferNode: true;
+
+    bufferType: string;
+    bufferCount: number;
+
+    constructor(value: ArrayLike<number>, bufferType: NodeTypeOption, bufferCount?: number);
```
joshuaellis, line 10: Isn't this class missing `getInputType` where the return type is only `"buffer"`?

types/three/examples/jsm/nodes/accessors/CameraNode.d.ts
```diff
@@ -0,0 +1,12 @@
+import Object3DNode, { Object3DNodeScope } from './Object3DNode';
+
+export type CameraNodeScope = Object3DNodeScope | typeof CameraNode.PROJECTION_MATRIX;
+
+export default class CameraNode extends Object3DNode {
+    static PROJECTION_MATRIX: 'projectionMatrix';
+
+    // @ts-expect-error
```
joshuaellis, line 8: I guess this is an issue because it's not the same time as it's parent class? 🤔 wonder if there's anything around `abstract` classes that could be helpful here.

types/three/examples/jsm/nodes/core/Node.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    nodeType: NodeTypeOption;
+    updateType: NodeUpdateTypeOption;
+    id: number;
+
+    constructor(nodeType?: NodeTypeOption);
+
+    getChildren(): Node[];
+    getHash(builder: NodeBuilder): string;
+    getUpdateType(builder: NodeBuilder): NodeUpdateTypeOption;
+    getNodeType(builder: NodeBuilder, output?: string | null): NodeTypeOption | null;
```
joshuaellis, line 18: IMO this isn't correct for any of the `builder` stuff, in this class specifically it's not there...

types/three/examples/jsm/nodes/accessors/UserDataNode.d.ts
```diff
@@ -0,0 +1,8 @@
+import { NodeTypeOption } from '../core/constants';
+import ReferenceNode from './ReferenceNode';
+
+export default class UserDataNode extends ReferenceNode {
+    userData: any;
+
+    constructor(property: string, inputType: NodeTypeOption, userData?: any);
```
joshuaellis: I believe userData should be an object

types/three/examples/jsm/nodes/core/CodeNode.d.ts
```diff
@@ -0,0 +1,12 @@
...
+import Node from './Node';
+import NodeBuilder from './NodeBuilder';
+
+export default class CodeNode extends Node {
+    isCodeNode: true;
+    code: string;
+    constructor(code: string, nodeType?: NodeTypeOption);
+
+    setIncludes(includes: any[]): this;
+    getIncludes(builder: NodeBuilder): any[];
```
joshuaellis: this doesn't actually require a bulder. I wonder if `unknown` would be better as opposed to `any` CC @Methuselah96 for his opinion.

types/three/examples/jsm/nodes/core/NodeBuilder.d.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export type BuildStageOption = 'construct' | 'analyze' | 'generate';
+
+export interface FlowData {
+    code: string;
+}
+
+export default class NodeBuilder {
+    object: Object3D;
+    material: Material;
```
joshuaellis, line 38: I don't know enough about `node` but from experience with things like `mesh` i'm wondering if a generic would be beneficial here to accurately depict the `Material` (same with BufferGeometry, Renderer & NodeParser)

types/three/examples/jsm/nodes/core/NodeBuilder.d.ts
```diff
@@ -0,0 +1,125 @@
...
+    removeStack(node: Node): void;
+    setHashNode(node: Node, hash: string): void;
+    addNode(node: Node): void;
+    getMethod(method: string): string;
+    getNodeFromHash(hash: string): Node;
+
+    addFlow(shaderStage: NodeShaderStageOption, node: Node): Node;
+
+    setContext(context: any): void;
+    getContext(): any;
```
joshuaellis: I think `context` is also supposed to be an object

types/three/examples/jsm/nodes/core/NodeParser.d.ts
```diff
@@ -0,0 +1,5 @@
+import Node from './Node';
+
+export default class NodeParser {
+    parseFunction(source: string): Node;
```
joshuaellis: Again, this doesn't actually take an argument and it doesn't return anything... 🤔

joshuaellis (changes requested):
Okay I don't want to overload this PR with feedback. 

Overall a few points are we should avoid `any` as much as humanly possible, inferring the type through reading more threejs stuff would be better (even if its wrong) because people _will_ fix it, if it's wrong (i know that sounds bad but i'm not suggesting we intentionally make it wrong).

Secondly a lot of the parent classes e.g. Node do not take the arguments they're defined as in here, the children classes do and we should be reflecting this. It might be that we have to use `abstract` to help with this (not totally sure).

Thirdly, it's still a great amount of work & a lot of it looks spot on, it's just about finessing the rough edges.

types/three/examples/jsm/nodes/core/Node.d.ts
miko3k: Yes. 

I came to a conclusion (maybe incorrect) that it's not really a "normal case" for `nodeType` to be `null` and if I included `null` in the result, any client code would have to deal with it.

I will change it though, I don't really have opinion here. 

edit: I just noticed that I defined the method `getNodeType` as `|null`. Kinda makes my previous sentence invalid. I will change it to include `null` in any case.

types/three/examples/jsm/nodes/core/Node.d.ts
miko3k: Can you elaborate (what method)?

types/three/examples/jsm/nodes/accessors/BufferNode.d.ts
miko3k: No, because it has a subclass `StorageBufferNode`. There are plenty of other nodes which call `super('vec3')` for instance. In some cases it might be possible to define `getNodeType`/`nodeType` to return this value, however it will also cause problems for possible inheritance later.

types/three/examples/jsm/nodes/accessors/CameraNode.d.ts
miko3k: This a general problem why I didn't want to be too specific in subclasses. In this case I decided to go this way, because I like the scope to be an union type, and I also hate to make `PROJECTION_MATRIX `an option for `Object3DNode `(also `typeof` would introduce circular depenency, I would have to be hardcoded). I concluded that `@ts-expect-error` is the most user friendly one. Alhough stricly speaking incorrect, when someone casts `CameraNode` to a parent.

edit: nothing can be really `abstract` either.

types/three/examples/jsm/nodes/core/Node.d.ts
miko3k: The subclasses require it. Not including it would break the inheritance. You cannot really override the method in subclass and add a parameter

(what nodes do in javascript is also correct, as including the parameter wouldn't change anything, javascript will happily call a method without specifying a parameter value)

types/three/examples/jsm/nodes/accessors/UserDataNode.d.ts
miko3k: Yes. I will change it.

types/three/examples/jsm/nodes/core/NodeBuilder.d.ts
miko3k: You are probably right, but I would opt for deleting the property from typings or just using a union type, `NodeBuilder` will not be used by majority anyways. And making it generic would genericize everything.  Honestly I also do not know nodes that well, I used my typing sucessully for a rather toy project before submitting them. I am open to suggestions here.

types/three/examples/jsm/nodes/core/NodeBuilder.d.ts
miko3k: Yes. I will change it.

types/three/examples/jsm/nodes/core/NodeParser.d.ts
miko3k: I believe the correct abstraction here is called `abstract` :) I will change it.

miko3k:
> Okay I don't want to overload this PR with feedback.

Feedback is definitely appreciated and it's a big change. I would like to reach the mergable state eventually. 

> Overall a few points are we should avoid `any` as much as humanly possible, inferring the type through reading more threejs stuff would be better (even if its wrong) because people _will_ fix it, if it's wrong (i know that sounds bad but i'm not suggesting we intentionally make it wrong).

This is a valid point. I will update the typings to reflect this. I generally used `any` as a synoynym for "object". Is there an agreement how to express an object in `@types/three`? `{}` or `object`? `Object3D` uses `{ [key: string]: any };` as `userData`. I would prefer `Record<string,unknown>`, I guess I will stick to to what `Object3D` does.

> 
> Secondly a lot of the parent classes e.g. Node do not take the arguments they're defined as in here, the children classes do and we should be reflecting this. It might be that we have to use `abstract` to help with this (not totally sure).
> 
I disagree with `abstract` as every node would have to override everything and it's very common for nodes not to simply fall back to default implementation. I explained in one of my previous commets why including the parameters is the correct choice. I think what I did is both relatively simple and correct form OO perspective.

edit: I will add `abstract` where nodes explictly say `console.warn( 'Abstract function.' );`. Almost. Bacause Node.update() is only used when nodeUpdateType !== none.

> Thirdly, it's still a great amount of work & a lot of it looks spot on, it's just about finessing the rough edges.

Thanks, I appreciate your recognition. I am sure there are bugs and the more can find now, the less breakage and anger for clients in the future.

types/three/examples/jsm/nodes/core/CodeNode.d.ts
miko3k: Was replaced by a new type `CodeNodeInclude`

miko3k:
Anyways. The current status

`any` was ditched... I introduced aliases `AnyJson`, `AnyObject` and `NodeUserData` in `constants.d.ts` so it's more manageable. The "object" and the "userData" are the same thing and they are compatible with `Object3D.userData`. `AnyJson` is still an alias for `any`.

Other uses `any` have been replaced by an actual type. I am quite proud of `LightsNode.setReference` alhough noone will ever use it :)

`abstract` has been sparingly introduced where appropriate. As I said earlier, `Node.update()` cannot be abstract, as it is required only if `updateType` is not `"none"`.

The point with inheritance and the respective types stays. Not sure how to proceed there. An example or closer description would be nice. No matter how I look at it, I cannot get the inheritance right.

joshuaellis:
Hey @miko3k sorry, I've been away the last week so will try to look at this for you over the next couple of days.

joshuaellis (approved):
I think realistically these types are better than no types and in regards to the inheritance it's a huge pain i know. That and the difficulty is we're writing types without changing the source to TS so i think realistically, we should introduce this (as some types are better than no types) and then we/people can refine them as we go forward.

So i'll merge this now and lets get it live...

miko3k:
Wow! Let's be brave I guess. I am pretty excited about this :)

Issue #35, opened 2021-02-28
[RFC] Should BufferAttribute.array be TypedArray instead of ArrayLike?


joshuaellis:
### Problem description:

Documented as BufferAttributes only being able to be given a TypedArray, but when you call `attribute.array` the return type is `ArrayLike` instead of a version of TypedArray – https://threejs.org/docs/#api/en/core/BufferAttribute

this becomes a problem in code like `estimateBytesUsed` in `BufferGeometryUtils`:

```ts
export function estimateBytesUsed(geometry: BufferGeometry) {
  let mem = 0
  for (let name in geometry.attributes) {
    const attr = geometry.getAttribute(name)
    mem += attr.count * attr.itemSize * (attr.array as TypedArray).BYTES_PER_ELEMENT // you have to typecase to get BYTES_PER_ELEMENT because it's not on ArrayLike
  }

  const indices = geometry.getIndex()
  mem += indices ? indices.count * indices.itemSize * (indices.array as TypedArray).BYTES_PER_ELEMENT : 0
  return mem
}
```

### Suggested solution:

Could we use a generic to detect the type of TypedArray passed to the Attribute so that type is returned when we call `attribute.array`?

dangerski:
This is also a problem when you try to update attribute values, which is shown in the threejs documentation.  ArrayLike is readonly so you can't change values using `[]`.

```
interface ArrayLike<T> {
    readonly length: number;
    readonly [n: number]: T;
}
```

[https://threejs.org/docs/#manual/en/introduction/How-to-update-things](https://threejs.org/docs/#manual/en/introduction/How-to-update-things)

positions will be readonly so this example from the documentation won't work in TS
```
const positions = line.geometry.attributes.position.array;

let x, y, z, index;
x = y = z = index = 0;

for ( let i = 0, l = MAX_POINTS; i < l; i ++ ) {
    positions[ index ++ ] = x;
    positions[ index ++ ] = y;
    positions[ index ++ ] = z;
    x += ( Math.random() - 0.5 ) * 30;
    y += ( Math.random() - 0.5 ) * 30;
    z += ( Math.random() - 0.5 ) * 30;

}

```

kmannislands:
For app code working with buffer geometries, it would also be nice to go a bit further and narrow the specific attributes that are defined as well as their respective types. This would require making BufferAttribute generic as well.

```typescript
const geom = new BufferGeometry<{ normal: Float32Array,  position: Uint16Array }>

[...]

PR #241, opened 2022-08-05
Draft: sketch out approach to better BufferAttribute typings #35

kmannislands:
This is a draft to collect feedback on the idea of improving specificity of BufferAttribute typings, especially as they relate to BufferGeometry typings

### Why

The current typing of attributes on BufferGeometries is too wide. It's not possible to specify what attributes are actually defined and what "flavor" of TypedArray actually underlies them.

### What

I've made the `BufferAttribute` type generic so that the backing TypedArrays can be specified. I've leveraged this in the definition of `BufferGeometry` to make it possible to specify the attributes and their underlying types.

With this change, the following code works:

```typescript
type geom = BufferGeometry<{ normal: Float32Array,  position: Uint16Array }>
type normal = geom['attributes']['normal'] // InterleavedBufferAttribute | BufferAttribute<Float32Array>
type position = geom['attributes']['position'] // InterleavedBufferAttribute | BufferAttribute<UInt16Array>

// @ts-expect-error Property 'foo' does not exist on type 'AttributesFor<{ normal: Float32Array; position: Uint16Array; }>'
type foo = geom['attributes']['foo'] // type error!
```

Using the previous wide type as default allows backwards compatibility:

```typescript
type geom = BufferGeometry
type foo = geom['attributes']['foo']; // InterleavedBufferAttribute | BufferAttribute<TypedArray>
```

If there's an appetite for this sort of change, I'll button up the PR here with: 
- Handling for all variants of TypedArray
- Handling of all methods of `BufferGeometry`
- Better handling/generic typing of `InterleavedBufferAttribute` in the same vein as the way `BufferAttribute` is handled

### Checklist

-   [ ] Checked the target branch (current goes `master`, next goes `dev`)
-   [ ] Ready to be merged

### DIFF
--- a/types/three/src/core/BufferAttribute.d.ts
+++ b/types/three/src/core/BufferAttribute.d.ts
@@ -5,14 +5,14 @@ import { Matrix4 } from './../math/Matrix4';
 /**
  * see {@link https://github.com/mrdoob/three.js/blob/master/src/core/BufferAttribute.js|src/core/BufferAttribute.js}
  */
-export class BufferAttribute {
-    constructor(array: ArrayLike<number>, itemSize: number, normalized?: boolean); // array parameter should be TypedArray.
+export class BufferAttribute<ArrayType extends ArrayLike<number> = ArrayLike<number>> {
+    constructor(array: ArrayType, itemSize: number, normalized?: boolean); // array parameter should be TypedArray.
 
     /**
      * @default ''
      */
     name: string;
-    array: ArrayLike<number>;
+    array: ArrayType;
     itemSize: number;
 
     /**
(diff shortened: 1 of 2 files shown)


Methuselah96:
I like where this is going. Could it be expanded to let the user say that an attribute is not a `GLBufferAttribute` (see https://github.com/three-types/three-ts-types/issues/389)?

Methuselah96:
Thanks for your work on this, sorry for the radio silence. I'm looking into building off of this to allow distinguishing between the different buffer attribute types (as discussed in https://github.com/three-types/three-ts-types/issues/389). Let me know if you would rather do that work yourself, and I'd be happy to review it.

PR #244, opened 2022-08-17
refactor: remove Quaternion.inverse()

### LINKED ISSUE #243: Quaternion.inverse() is marked as deprecated instead of removed.
-   `three` version: 0.143.0
-   `node` version: v18.5.0
-   `npm` (or `yarn`) version: 8.12.1

### Problem description:

Quaternion.inverse() is marked as deprecated instead of removed.

### Relevant code:

```ts
const quat = new Quaternion();
quat.inverse()
```

`.inverse` is marked as "deprecated". 
https://github.com/three-types/three-ts-types/blob/master/types/three/src/math/Quaternion.d.ts#L174

Editors like VSCode will show a strike-through for such methods, but usually this indicates a method that *actually exists*. Otherwise, this causes unexpected runtime errors that the static features of TypeScript are meant to solve, like: https://github.com/cubing/cubing.js/issues/195

### Suggested solution:

Remove this line:

https://github.com/three-types/three-ts-types/blob/master/types/three/src/math/Quaternion.d.ts#L174

Possibly also some other methods, if they are also gone?

lgarron:
### Why

Fix:

- #243 

### What

Remove a method that's gone.

### Checklist

-   [ ] Ready to be merged

### DIFF
--- a/types/three/src/math/Quaternion.d.ts
+++ b/types/three/src/math/Quaternion.d.ts
@@ -168,11 +168,6 @@ export class Quaternion {
      */
     multiplyVector3(v: any): any;
 
-    /**
-     * @deprecated Use {@link Quaternion#invert .invert()} instead.
-     */
-    inverse(): Quaternion;
-
     random(): Quaternion;
 
     [Symbol.iterator](): Generator<number, void>;


joshuaellis (approved):

joshuaellis:
Thanks for the contribution!

Issue #239, opened 2022-07-28
r144


joshuaellis:
### Issue Tracker

This is used to track potential issues we need to investigate surrounding the `r144` release.

NOTE: i've done quite a few already on `dev` branch, these are the remaining from the last checked

- [x] https://github.com/mrdoob/three.js/issues/24453
- [x] https://github.com/mrdoob/three.js/issues/24495
- [x] https://github.com/mrdoob/three.js/issues/24504

Last checked: mrdoob/three.js@a6cee1f74dd36f4277440597c1b08483e2a79595 (31 / 08 / 22).

joshuaellis:
@miko3k would you be interested in taking a look at https://github.com/mrdoob/three.js/pull/24504, took a go myself but there was inheritance issues with Node vs ShaderNode. No worries if not 👍🏼 work off the `dev` branch though!

joshuaellis:
will be resolved by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62071

PR #245, opened 2022-08-21
feat: fn/func in nodes

miko3k:
### Why

#239 

### What

**please note that this PR also changes `package.json` a bit to enable both `yarn test` and `yarn test-all` to work**

- bringing relevant typing to sync with r144
- general improvements of `FunctionNode`/`FunctionCallNode` with generics modelling the required parameter
   -  note that `func` returning `ShaderNode` was a hack, in reality it was something  `ShaderNode`-like (having `call`-method accepting `Swizzable`). `FunctionNode` as the generic parameter was plainly wrong.
- automatically generate proxified versions of parameters of `FunctionNode` using mapped type `Fn`
- `fn`/`func` take same generic argument as `FunctionNode`
- `MakeObjectOption` was renamed to `Proxied`, I added some comments although I am still not too happy
- relevant tests
- newly defined types:
  - `FunctionNodeArguments`, pretty self-explanatory
  - `Fn` converts `FunctionNodeArguments` into `Proxied` version of it

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
Great work!

PR #256, opened 2022-09-02
refactor: Object3D.matrixWorldAutoUpdate, deprecate Scene.autoUpdate

CodyJasonBennett:
### Why

Follow-up to https://github.com/mrdoob/three.js/pull/24028 where `Scene.autoUpdate` is moved to `Object3D.matrixWorldAutoUpdate` for granular control of matrices on render and via matrix methods.

### What

I've added the new `Object3D` properties and deprecated `Scene.autoUpdate`. They added the latter to `Three.Legacy` which will be removed entirely in 10 releases (should this info be inlined?).

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/core/Object3D.d.ts
+++ b/types/three/src/core/Object3D.d.ts
@@ -110,6 +110,13 @@ export class Object3D<E extends BaseEvent = Event> extends EventDispatcher<E> {
     matrixAutoUpdate: boolean;
 
     /**
+     * When this is set, the renderer checks every frame if the object and its children need matrix updates.
+     * Otherwise, you have to maintain all matrices in the object and its children yourself.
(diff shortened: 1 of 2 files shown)


joshuaellis:
> They added the latter to Three.Legacy which will be removed entirely in 10 releases (should this info be inlined?).

We don't normally actively deprecate things, we normally just remove them completely. If you can do that, that would be great. I do appreciate the method exists but iirc it just logs to the console?

CodyJasonBennett:
The property exists as an alias via a getter/setter but will emit a warning regarding the rename. They do this for all deprecations, including the recent *BufferGeometry deprecations.

joshuaellis (approved):

PR #257, opened 2022-09-02
fix(ColorManagement): export as namespace, add static properties

### LINKED ISSUE #250: Missing definition for THREE.ColorManagement.legacyMode
-   `three` version: 0.143.0
-   `node` version: v18.7.0
-   `npm` (or `yarn`) version: 1.22.10 (yarn)

### Problem description:

I  have run into a few issues regarding the `ColorManagement` module:
- `ColorManagement` has a `legacyMode` property that is missing from the type definitions. Here's the property in the latest release of `three` (`three@0.143.0`): [link](https://github.com/mrdoob/three.js/blob/e22cb060cc91283d250e704f886528e1be593f45/src/math/ColorManagement.js#L24).
- `ColorManagement` is not exported properly. See how it's exported in  [`threejs`](https://github.com/mrdoob/three.js/blob/e22cb060cc91283d250e704f886528e1be593f45/src/Three.js#L131), but not in [`@types/three`](https://github.com/three-types/three-ts-types/blob/7d43269b9e7c482bed5d2a0f0e16b5c00ab70a8d/types/three/src/Three.d.ts#L135).

I'd be happy to make a PR if that helps. :relaxed: 

### Relevant code:j

```ts
import * as THREE from 'three';
// The following lines result in type errors, but they
// don't throw any errors when they are actually executed
THREE.ColorManagement;
THREE.ColorManagement.legacyMode;
```

### Suggested solution:

I'd love to add the missing type definitions myself in a PR, if you're okay with that. :relaxed:

CodyJasonBennett:
### Why

Support for globally correct color management was added to r139 with https://github.com/mrdoob/three.js/pull/23392. Background and API usage can be found at https://threejs.org/docs/#manual/en/introduction/Color-management.

### What

I've added missing static properties and exported `ColorManagement`, its `SRGBToLinear` and `LinearToSRGB` methods are private and not user-facing.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -158,6 +158,7 @@ export * from './math/Vector2';
 export * from './math/Quaternion';
 export * from './math/Color';
 export * from './math/SphericalHarmonics3';
+export { ColorManagement } from './math/ColorManagement'
 import * as MathUtils from './math/MathUtils';
 export { MathUtils };
 /**
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -158,7 +158,7 @@ export * from './math/Vector2';
 export * from './math/Quaternion';
 export * from './math/Color';
 export * from './math/SphericalHarmonics3';
-export { ColorManagement } from './math/ColorManagement'
+export { ColorManagement } from './math/ColorManagement';
 import * as MathUtils from './math/MathUtils';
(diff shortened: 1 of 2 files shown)


joshuaellis (approved):
Thanks!

joshuaellis:
Looks like you need to run `prettier`?

PR #258, opened 2022-09-06
fix: Remove url argument from PCDLoader.parse signature

grschafer:
### Why

The `url` argument to `PCDLoader` was removed from three.js in r141 (May 2022): https://github.com/mrdoob/three.js/commit/eab2a29621b332e0e8a0bac5a0098cfcf1e331e1#diff-d7eb86747ff8ee3e2ab047235419fa6870b9f13e6a0be07a7a35e92064f64b86R56

### What

This PR updates the types to match the current implementation of PCDLoader in three.js.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -386,6 +386,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "grschafer",
+      "name": "Greg Schafer",
+      "avatar_url": "https://avatars.githubusercontent.com/u/694225?v=4",
+      "profile": "https://github.com/grschafer",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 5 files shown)


joshuaellis (changes requested):
Is it possible to add a test for this loader? Perhaps we can copy the example it's used in from `three`?

grschafer:
Added a test which is mostly a copy of the three.js PCDLoader example:
- https://github.com/mrdoob/three.js/blob/68daccedef9c9c325cc5f4c929fcaf05229aa1b3/examples/webgl_loader_pcd.html

Unfortunately, that example doesn't exercise the `PCDLoader.parse` function, which is the one that needed its types updated! I looked around for another PCDLoader example that exercises `parse`, but the only one in three.js itself is in [the editor project](https://github.com/mrdoob/three.js/blob/68daccedef9c9c325cc5f4c929fcaf05229aa1b3/editor/js/Loader.js#L522), which contains a lot of other loaders and some editor-specific relative imports.

So instead, I just copied the [load function from the source of `PCDLoader`](https://github.com/mrdoob/three.js/blob/68daccedef9c9c325cc5f4c929fcaf05229aa1b3/examples/jsm/loaders/PCDLoader.js) (see the `pcdLoaderLoad` function) into the PCDLoader example and slightly modified it to compile.

Test output without and with the `fix: Remove url arg from PCDLoader.parse signature` commit:

```
$ npm run test

> test
> node node_modules/@definitelytyped/dtslint/dist/index.js types/three

dtslint@0.0.111
Error: /home/greg/repos/three-ts-types/types/three/test/loaders/loaders-pcdloader.ts:75:30
ERROR: 75:30  expect  TypeScript@4.7 compile error: 
Expected 2 arguments, but got 1.

    at testTypesVersion (/home/greg/repos/three-ts-types/node_modules/@definitelytyped/dtslint/dist/index.js:194:15)
    at async runTests (/home/greg/repos/three-ts-types/node_modules/@definitelytyped/dtslint/dist/index.js:170:13)
    at async main (/home/greg/repos/three-ts-types/node_modules/@definitelytyped/dtslint/dist/index.js:79:9)

$ npm run test

> test
> node node_modules/@definitelytyped/dtslint/dist/index.js types/three

dtslint@0.0.111
```

If this approach (manual modification of a three.js example) is a bad idea, please let me know what a preferable approach would be! Thanks :smile:

joshuaellis:
Works for me, I'll review when at my computer 👌🏻 thanks for doing that.

joshuaellis (approved):

PR #261, opened 2022-09-09
fix(Mesh): material default is MeshBasicMaterial

CodyJasonBennett:
### Why

`THREE.Mesh` is created with an empty `THREE.BufferGeometry` and `THREE.MeshBasicMaterial` by default, but its material type defaults to `Material | Material[]`.

### What

I've changed `Mesh`'s default material to `MeshBasicMaterial` to better reflect its implementation. This shouldn't change inference when geometry/material are defined as args, but I worry about the case where they are directly mutated and if that's the reason for this discrepancy in the first place.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/objects/Mesh.d.ts
+++ b/types/three/src/objects/Mesh.d.ts
@@ -2,11 +2,12 @@ import { Material } from './../materials/Material';
 import { Raycaster } from './../core/Raycaster';
 import { Object3D } from './../core/Object3D';
 import { BufferGeometry } from '../core/BufferGeometry';
+import { MeshBasicMaterial } from '../materials/MeshBasicMaterial';
 import { Intersection } from '../core/Raycaster';
 
 export class Mesh<
     TGeometry extends BufferGeometry = BufferGeometry,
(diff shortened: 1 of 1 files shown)


joshuaellis:
Yep, I agree this should certainly be the default considering https://github.com/mrdoob/three.js/blob/2a7005f1a1df6d0e7b64e60ae08bff3f9ca41c17/src/objects/Mesh.js#L37, Just need the tests to be fixed before we can merge this! 😄

CodyJasonBennett:
Actually, I think this would be a huge regression for those typing fields as `{ param: Mesh }` etc. since this PR incidentally increases that parameter's specificity -- it wouldn't be valid to pass any arbitrary mesh there anymore. This is also very evident when mutating materials to swap between renders. In both cases, you'd have to revert the materials to `THREE.Mesh<THREE.BufferGeometry, THREE.Material | THREE.Material>`.

Closing this as a precaution for the first case, specifically. Tests would be available in https://github.com/CodyJasonBennett/three-ts-types/commit/fdcf817acbb319700ad714c6b410271bfe07e3d3.

PR #269, opened 2022-10-07
fix: Change `getObjectByProperty` value to `any`

mattrossman:
### Why

The Three.js docs list the `value` for getObjectByProperty as "Any" type:

https://threejs.org/docs/?q=mesh#api/en/core/Object3D.getObjectByProperty

Currently the type is defined as `string` which creates errors for non-string values:

```ts
bone = root.getObjectByProperty("isBone", true)
```

### What

`string` -> `any`

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):
If you can resolve the conflicts, this can be merged. Thanks for the contribution!

mattrossman:
Updated to resolve conflict

PR #295, opened 2022-12-15
[feat] allow GLBufferAttribute as an attribute type

mikkom:
### Why

The current typings don't allow setting `GLBufferAttribute` as an attribute in `BufferGeometry`. This functionality has been in Three.js for quite some time (originally introduced in https://github.com/mrdoob/three.js/pull/13196 back in 2020).

Three.js master continues to support this feature in `BufferGeometry` (see e.g. https://github.com/mrdoob/three.js/blob/master/src/core/BufferGeometry.js#L371), documents the attribute type in https://threejs.org/docs/#api/en/core/GLBufferAttribute and has a usage example in https://github.com/mrdoob/three.js/blob/master/examples/webgl_buffergeometry_glbufferattribute.html#L115-L117.

### What

This change adds `GLBufferAttribute` as a valid type for `setAttribute` and `getAttribute` methods in `BufferGeometry`. The type for `GLBufferAttribute` had already been defined and exported but it was not used anywhere else in the typings.

Note that the change can break consumers of the typings as `GLBufferAttribute` does not implement the same interface as `BufferAttribute`. The way to fix these issues is to use the boolean flags (`isBufferAttribute`, `isInterleavedBufferAttribute`, or `isGLBufferAttribute`) to discriminate between the different options (or alternatively by resorting to type assertions). Despite the potential breakage, I believe these typings are more correct.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/examples/jsm/nodes/shadernode/ShaderNodeBaseElements.d.ts
+++ b/types/three/examples/jsm/nodes/shadernode/ShaderNodeBaseElements.d.ts
@@ -115,17 +115,17 @@ export type Fn<P extends FunctionNodeArguments> = P extends readonly [...unknown
     ? ProxiedTuple<P>
     : readonly [ProxiedObject<P>];
 
-// tslint:disable:no-unnecessary-generics
 export function func<P extends FunctionNodeArguments>(
     code: string,
     includes?: CodeNodeInclude[],
+    // eslint-disable-next-line no-unnecessary-generics
 ): { call: (...params: Fn<P>) => Swizzable };
 
 export function fn<P extends FunctionNodeArguments>(
     code: string,
     includes?: CodeNodeInclude[],
+    // eslint-disable-next-line no-unnecessary-generics
 ): (...params: Fn<P>) => Swizzable;
-// tslint:enable:no-unnecessary-generics
 
 export const instanceIndex: Swizzable<InstanceIndexNode>;
 export function label(node: NodeRepresentation, name?: string): Swizzable<VarNode>;
--- a/types/three/src/loaders/ObjectLoader.d.ts
+++ b/types/three/src/loaders/ObjectLoader.d.ts
@@ -13,8 +13,7 @@ export class ObjectLoader extends Loader {
 
     load(
         url: string,
-        // tslint:disable-next-line:no-unnecessary-generics
-        onLoad?: <ObjectType extends Object3D>(object: ObjectType) => void,
+        onLoad?: (object: Object3D) => void,
         onProgress?: (event: ProgressEvent) => void,
         onError?: (event: Error | ErrorEvent) => void,
     ): void;
--- a/types/three/test/nodes/nodes-FunctionNode.ts
+++ b/types/three/test/nodes/nodes-FunctionNode.ts
@@ -22,7 +22,7 @@ const includes = [mx_noise];
 const someFunc1 = new FunctionNode<[a: Node]>();
 const someFunc2 = new FunctionNode<{ a: Node }>();
 
-// tslint:disable-next-line:no-unnecessary-generics
+// eslint-disable-next-line no-unnecessary-generics
 function assertSwizzable<T extends Node>(_s: Swizzable<T>) {}
 
 type a = ProxiedObject<readonly [Node]>;
--- a/types/three/test/nodes/nodes-ShaderNode.ts
+++ b/types/three/test/nodes/nodes-ShaderNode.ts
@@ -20,7 +20,7 @@ import {
 import { ConvertType, Swizzable } from 'three/examples/jsm/nodes/shadernode/ShaderNode';
 
 // just to type check
-// tslint:disable-next-line:no-unnecessary-generics
+// eslint-disable-next-line no-unnecessary-generics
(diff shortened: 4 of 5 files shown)


types/three/src/core/BufferAttribute.d.ts
```diff
@@ -43,6 +43,8 @@ export class BufferAttribute {
     set needsUpdate(value: boolean);
 
     readonly isBufferAttribute: true;
+    readonly isInterleavedBufferAttribute: false;
+    readonly isGLBufferAttribute: false;
```
0b5vr: I don't think these `false` are explicitly set.

types/three/src/core/GLBufferAttribute.d.ts
```diff
@@ -5,13 +5,19 @@
 export class GLBufferAttribute {
     constructor(buffer: WebGLBuffer, type: number, itemSize: number, elementSize: 1 | 2 | 4, count: number);
 
+    /**
+     * @default ''
+     */
+    name: string;
```
0b5vr: where is it set?

0b5vr (changes requested):
Check 2 comments I left. Otherwise it looks good to me.

types/three/src/core/GLBufferAttribute.d.ts
mikkom: Good point. For whatever reason the name property isn't initialized as an empty string in `GLBufferAttribute` even though both the other attribute types do that. I'll remove the comment and the property (even though this adds risk of further breakage if people are relying on using the name property on attributes...).

types/three/src/core/BufferAttribute.d.ts
mikkom: True that. Should I rather set these as `undefined` instead of `false`? If we omit these fields completely then we no longer can use them to discriminate between the different attribute types.

types/three/src/core/GLBufferAttribute.d.ts
mikkom: Another option would be to set the field as
```
    /**
     * @default undefined
     */
    name?: string;
```

types/three/src/core/BufferAttribute.d.ts
0b5vr: Ok that's the motivation, I got it.
You want to use `position.isGLBufferAttribute` here right?

https://github.com/three-types/three-ts-types/blob/9ce3aa43d8b9af2d63217cd9dfa177d7ae4edcbe/types/three/test/controls/controls-pointercontrols.ts#L129

But having `false` here only for this is lying since it isn't actually `false`.

You can solve this by either of these 3 solutions:

1. use `!(position instanceof GLBufferAttribute)` instead for the condition
2. use `'setXYZ' in position` instead for the condition
3. type assert the `position`, `let position = floorGeometry.attributes.position as BufferAttribute;`

I prefer the (3), since it's obvious that the `position` is `BufferAttribute` here since it's made by `new THREE.PlaneGeometry()`.

types/three/src/core/GLBufferAttribute.d.ts
0b5vr: I agree that it's weird `GLBufferAttribute` doesn't have `name` on the Three.js side.
But, if it's not appeared in the Three.js source yet, it should not be added here I think.

> (even though this adds risk of further breakage if people are relying on using the name property on attributes...)

That's actually a good point.
I feel like we have to ask Three.js side about `GLBufferAttribute.name` first.

> Another option would be to set the field as
> 
> ```
>     /**
>      * @default undefined
>      */
>     name?: string;
> ```

I choose this if Three.js side says `GLBufferAttribute.name` should exist.

@mikkom Would you ask about this somewhere? I would choose [Discourse](https://discourse.threejs.org/).

0b5vr:
You did it on Three.js side, wooo! Thank you!

types/three/src/core/GLBufferAttribute.d.ts
0b5vr: You did it on Three.js side, wooo! Thank you! Now it's valid to say the `.name` (in the next revision, r149).

mikkom: Heh yeah, finally got to pushing this forward :)  I think I should now retarget this PR against dev instead of master, right?

types/three/src/core/BufferAttribute.d.ts
mikkom: To me it seems that it would be most natural to use the publicly documented boolean fields (e.g. https://threejs.org/docs/#api/en/core/BufferAttribute.isBufferAttribute) for checking this. So my suggestion would be to still  change the `false` values to `undefined` and then the types would precisely match the returned values. If we omit the non-true fields from TS types then it's not possible to refer to the boolean field unless you already have the corresponding attribute type. And then there's no longer any point in checking a trivially true field :)

The three solutions you listed above definitely do work and nothing prevents the user from choosing any of them. And I agree that in many cases a type assertion can be the most straightforward option. I just think that it would be better to support the documented fields in a meaningful way in TS, as well. WDYT?

PR #298, opened 2022-12-18
`ObjectLoader.parseImages`: onLoad is optional, returns a record of `…

Hoodgail:


joshuaellis:
Going to need you to explain this change please.

Hoodgail:
https://github.com/mrdoob/three.js/blob/dev/src/loaders/ObjectLoader.js#L363

Here, an empty object is defined, named `images`
image
-------------------
Here is where the sources are assigned to the empty object by their uuid
https://github.com/mrdoob/three.js/blob/dev/src/loaders/ObjectLoader.js#L459
https://github.com/mrdoob/three.js/blob/dev/src/loaders/ObjectLoader.js#L466
image

And then returned
image

Hoodgail:
As for `onLoad`
It gets sent to LoadingManager
image
--------------------------
And here you can see that onLoad can be optional 
https://github.com/mrdoob/three.js/blob/362228a5a6d961b547accc2048ce9bfb8c2875fd/src/loaders/LoadingManager.js#L53
--------------------------
Here is an example of onLoad not being used.
https://github.com/mrdoob/three.js/blob/dev/editor/js/commands/SetMaterialCommand.js#L64
image

0b5vr (approved):

PR #299, opened 2022-12-18
`Object3D`: class instances

Hoodgail:


joshuaellis:
CI failing.

Hoodgail:
> CI failing.

image
No idea why

joshuaellis:
> > CI failing.
> 
> image No idea why

I would look at the test files that are failing and try to address from there.

0b5vr:
What is the motivation for this change? I thought the use of `this` here is helpful on subclasses.

Hoodgail:
> What is the motivation for this change? I thought the use of `this` here is helpful on subclasses.
@0b5vr 
I made some more reviewing and i think you're right, i don't remember the error i was getting with `:this`

But now i wonder why the others don't do the same like `BufferGeometry`

image
image
image

0b5vr:
mmm yes, I think `BoxGeometry.clone` should return `BoxGeometry`.

PR #300, opened 2022-12-18
`ObjectLoader.parseMaterials`: accepts a record of textures not an ar…

Hoodgail:


joshuaellis:
Going to need you to explain this change please.

Hoodgail:
@joshuaellis 

1.  the textures argument gets sent to `MaterialLoader.setTextures`
https://github.com/mrdoob/three.js/blob/dev/src/loaders/ObjectLoader.js#L311
image
------------------
2.  Set textures simply assings it to the class as a property
https://github.com/mrdoob/three.js/blob/dev/src/loaders/MaterialLoader.js#L319
image
------------------
3.  But when using them, its used as an object.
https://github.com/mrdoob/three.js/blob/dev/src/loaders/MaterialLoader.js#L73
image
------------------
The name argument in getTexture is not a number, its a string. The uuids of the textures
image

## Sorry for the late reply.

0b5vr:
seems the doc is outdated...

https://threejs.org/docs/#api/en/loaders/ObjectLoader.parseMaterials

Is this a required parameter?

0b5vr (approved):

Hoodgail:
> seems the doc is outdated...
> 
> https://threejs.org/docs/#api/en/loaders/ObjectLoader.parseMaterials
> 
> Is this a required parameter?

image
@0b5vr 
I'm not sure, but if it were optional, there would be an if statement for determining whether the parameter is undefined or not; since there doesn't seem to be one, i'd say it's required. However, an empty object can be accepted because it checks to see if a requested value in the object is undefined.

0b5vr:
Yeah it seems to be.

Issue #294, opened 2022-12-15
Does KeyframeTracks parameter times, values have to be any type?


Suprhimp:
Hi, I faced the runtime error because of my keyframe value has Nan value Because I put the wrong type of the input to make keyframe.

In `THREE.VectorKeyframeTrack` the paramter `times` and `values` have to be number array.

First I made the code like this.
```
posKF = new THREE.VectorKeyframeTrack('.position',timesList, [camera.position], THREE.InterpolateSmooth);
```
But as you know I missed `.toArray()` function for camera position value to array of number not object.
```
posKF = new THREE.VectorKeyframeTrack('.position',timesList, [...camera.position.toArray()], THREE.InterpolateSmooth);
```

When I used the code with first way It compiled but I have to check `posKF` value because of not working well.

I logged `posKF` , it gives me this
![image](https://user-images.githubusercontent.com/73486185/207858507-a3c4a7b3-55e2-430e-8b68-03f6b7691758.png)

The values of keyframe has `Nan` value.

I think  `VectorKeyframeTrack`'s  parameter `times` and `values` have to use type `number[]` instead of `any[]`.

And so on `KeyframeTrack`'s parameter `times` and `values` have to use type `ArrayLike<number>` instead of `ArrayLike<any>` .

Does I'm thinking wrong? If it is not, I want to make PR for that.

Thanks for read :)

0b5vr:
Docs: https://threejs.org/docs/#api/en/animation/KeyframeTrack
Implementation: https://github.com/mrdoob/three.js/blob/309b00afb6dcbc5e6c58e72f10eaa8d2e8888c83/src/animation/KeyframeTrack.js#L20-L21
`AnimationUtils.convertArray`: https://github.com/mrdoob/three.js/blob/309b00afb6dcbc5e6c58e72f10eaa8d2e8888c83/src/animation/AnimationUtils.js#L19-L33

is it `ArrayLike<number>` ? Low confidence

Suprhimp:
As document saids
```
times - an array of keyframe times, converted internally to a Float32Array
values - an array with the values related to the times array, converted internally to a Float32Array.
```

My opinion is `time` and `value` has to be number convert value to `Float32`

Shall I ask in three js github issue?

0b5vr:
Issue is too much for ask this imo. I just asked in Discord

0b5vr:
donmccurdy says that `ArrayLike<number>` looks correct.

[...]

PR #306, opened 2022-12-23
Change KeyframeTrack params type `times` from any[] to number[]

Suprhimp:
Change KeyframeTrack params `times` from `any[]` to `number[]`

By doing this following child of KeyframeTrack, I changed `BooleanKeyframeTrack` , `ColorKeyframeTrack`, `NumberKeyframeTrack ` ,`StringKeyframeTrack ` , `QuaternionKeyframeTrack ` , `VectorKeyframeTrack `.

I changed `times` types of these. And I changed `QuaternionKeyframeTrack ` , `VectorKeyframeTrack ` parameter `values` from `any[]` to `number[]`.

But low confident, I didn't changed other `values` types.
I think it will be good if we can know what others type of the value and change by this PR.

If I know the type of other values, I will add new commit.

-   [x ] Checked the target branch (current goes `master`, next goes `dev`)
-   [ x] Ready to be merged

I didn't add my self to contributors table cause it already has.

issue link https://github.com/three-types/three-ts-types/issues/294

### DIFF
--- a/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
+++ b/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
@@ -2,7 +2,7 @@ import { KeyframeTrack } from './../KeyframeTrack';
 import { InterpolationModes } from '../../constants';
 
 export class NumberKeyframeTrack extends KeyframeTrack {
-    constructor(name: string, times: number[], values: any[], interpolation?: InterpolationModes);
+    constructor(name: string, times: ArrayLike<number>, values: ArrayLike<number>, interpolation?: InterpolationModes);
 
     /**
      * @default 'number'
--- a/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
+++ b/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
@@ -2,7 +2,7 @@ import { KeyframeTrack } from './../KeyframeTrack';
 import { InterpolationModes } from '../../constants';
 
 export class NumberKeyframeTrack extends KeyframeTrack {
-    constructor(name: string, times: any[], values: any[], interpolation?: InterpolationModes);
+    constructor(name: string, times: number[], values: any[], interpolation?: InterpolationModes);
 
     /**
      * @default 'number'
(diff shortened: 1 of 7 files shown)


types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
```diff
@@ -2,7 +2,7 @@ import { KeyframeTrack } from './../KeyframeTrack';
 import { InterpolationModes } from '../../constants';
 
 export class NumberKeyframeTrack extends KeyframeTrack {
-    constructor(name: string, times: any[], values: any[], interpolation?: InterpolationModes);
+    constructor(name: string, times: number[], values: any[], interpolation?: InterpolationModes);
```
0b5vr: oo, `ArrayLike<number>` and `number[]` are different since `number[]` does not accept `Float32Array`.

This is the definition of `ArrayLike<T>`:
https://github.com/microsoft/TypeScript/blob/acf854b636e0b8e5a12c3f9951d4edfa0fa73bcd/lib/lib.es5.d.ts#L1554-L1557

I believe `values` is also `ArrayLike<number>`.
`BooleanKeyframeTrack` and `StringKeyframeTrack` are exceptions, but otherwise values can be `Float32Array` too.

(and I don't know why it used to be `any[]` which does not accept `Float32Array` ...)

0b5vr (changes requested):

types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
Suprhimp: Okay than,
All `times` params have to be `ArrayLike<number>` and, `values` have to be `ArrayLike<number>` except `BooleanKeyframeTrack` and `StringKeyframeTrack` right?

(Any way can I ask what is oo means?)

0b5vr: yeah, I think that's right.

> (Any way can I ask what is oo means?)

Just an interjection 😄

Suprhimp: I revised it! 

I hope you've been a great holiday and happy new year :)

0b5vr (approved):

0b5vr:
Thanks!

PR #313, opened 2023-01-06
fix: move toHalfFloat/fromHalfFloat into DataUtils namespace

rotu:
### Why

Previously, trying to `import {toHalfFloat, fromHalfFloat} from 'three'` would typecheck but would fail when run as JavaScript. `import {DataUtils} from 'three'` would fail to typecheck but `DataUtils.toHalfFloat` would run as JavaScript

### What

I have updated the types to match their declaration.

### Checklist

-   [x] Ready to be merged


0b5vr:
The blame says this: https://github.com/three-types/three-ts-types/commit/59f15d614bd844e199be5326659c77c35a9dd038

Probably @joshuaellis forgot to change `Three.d.ts` side?

https://github.com/three-types/three-ts-types/blob/3ad9711aa8d61f9a5ea0c1a6a89d66c65fc46d84/types/three/src/Three.d.ts#L69

I think we should change the line to:

```js
export * as DataUtils from './extras/DataUtils';
```

0b5vr:
thank you @rotu .
awaiting @joshuaellis 's reponse, it's what you intended right?

joshuaellis (approved):
Yep, looks like a mistake. Great catch.

PR #314, opened 2023-01-06
[feat] allow GLBufferAttribute as an attribute type

mikkom:
This is the same change as https://github.com/three-types/three-ts-types/pull/295, just retargeted against dev branch.

### Why

The current typings don't allow setting `GLBufferAttribute` as an attribute in `BufferGeometry`. This functionality has been in Three.js for quite some time (originally introduced in https://github.com/mrdoob/three.js/pull/13196 back in 2020).

Three.js master continues to support this feature in `BufferGeometry` (see e.g. https://github.com/mrdoob/three.js/blob/master/src/core/BufferGeometry.js#L371), documents the attribute type in https://threejs.org/docs/#api/en/core/GLBufferAttribute and has a usage example in https://github.com/mrdoob/three.js/blob/master/examples/webgl_buffergeometry_glbufferattribute.html#L115-L117.

### What

This change adds `GLBufferAttribute` as a valid type for `setAttribute` and `getAttribute` methods in `BufferGeometry`. The type for `GLBufferAttribute` had already been defined and exported but it was not used anywhere else in the typings.

Note that the change can break consumers of the typings as `GLBufferAttribute` does not implement the same interface as `BufferAttribute`. The way to fix these issues is to use the boolean flags (`isBufferAttribute`, `isInterleavedBufferAttribute`, or `isGLBufferAttribute`) to discriminate between the different options (or alternatively by resorting to type assertions). Despite the potential breakage, I believe these typings are more correct.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -504,6 +504,15 @@
         "bug",
         "code"
       ]
+    },
+    {
+      "login": "mikkom",
+      "name": "Mikko Matilainen",
+      "avatar_url": "https://avatars.githubusercontent.com/u/77391?v=4",
+      "profile": "https://github.com/mikkom",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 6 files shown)


0b5vr:
oh, thanks for retargeting the target branch

types/three/src/core/BufferAttribute.d.ts
```diff
@@ -61,6 +61,8 @@ export class BufferAttribute {
     set needsUpdate(value: boolean);
 
     readonly isBufferAttribute: true;
+    readonly isInterleavedBufferAttribute: undefined;
+    readonly isGLBufferAttribute: undefined;
```
0b5vr: if it's undefined, we don't have to explicitly say this I think.

0b5vr: ah, I just saw the message you left in the previous PR. https://github.com/three-types/three-ts-types/pull/295#discussion_r1062246152

If it's the solution I feel like we have to set `isXXX` to every Three.js class, and I don't think it's the right way.
(e.g. saying `readonly isMesh: undefined` in `Matrix4` because someone might want to use that)
I believe people solve this kind of problems in other ways in the TypeScript realm.

mikkom: Sorry about the delayed response. I think you make a fair point there. It could indeed get unwieldy quite fast if we would need to cover all the fields in similar fashion. So I've now removed the explicit `undefined` field definitions and switched to using `instanceof` in the pointercontrols test.

0b5vr (approved):
Now it looks good, Thanks!

PR #315, opened 2023-01-09
feat: add generic arg for uniform class

JoshStern:
### Why

The uniform class directly wraps another value and it can be useful to annotate what that value is.

### What

Adds a generic argument for the uniform class. The default is left as `any` avoid making a breaking change.

### Checklist

-   [x] Ready to be merged


joshuaellis (approved):

0b5vr:
note: while THREE.Uniform is a proper thing, the format we use for uniforms is usually just an object with a `value` and that's called `IUniform` in three-ts-types
https://github.com/three-types/three-ts-types/blob/3ad9711aa8d61f9a5ea0c1a6a89d66c65fc46d84/types/three/src/renderers/shaders/UniformsLib.d.ts#L1-L4

0b5vr (approved):

PR #318, opened 2023-01-13
fix: allow null for WebXRManager.setSession

saitonakamura:
### Why

setting session as null doesn't really do anything, but it's used in this code https://github.com/mrdoob/three.js/blob/312ee7b016c9bb35abe2b9e627829a7d4eb89a2e/editor/js/Viewport.VR.js#L60 so I guess it's fine to allow it

### What

 allow null for WebXRManager.setSession

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -656,6 +656,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "saitonakamura",
+      "name": "Michael サイトー 中村 Bashurov",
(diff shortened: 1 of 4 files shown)


0b5vr:
Saw the three.js source code. It is probably for telling the JS engine that the xr session is no longer used?

0b5vr (approved):

0b5vr:
@saitonakamura it seems there are conflicts. could you fix this? :bow:

saitonakamura:
sure thing

PR #350, opened 2023-02-08
check color names

rotu:
### Why

`Color`'s constructor previously allowed invalid color names, which fail at runtime with a warning.

### What

Now colornames will be checked by typescript and will autocomplete in supporting ides.

E.g. `new Color('blu')` will fail to typecheck.

### Checklist

-   [X] Ready to be merged


types/three/src/math/Color.d.ts
```diff
@@ -5,6 +5,159 @@ import { BufferAttribute } from './../core/BufferAttribute';
 
 export { SRGBToLinear } from './ColorManagement';
 
+declare const _colorKeywords: {
```
0b5vr, line 8: source: https://github.com/mrdoob/three.js/blob/b9bc47ab1978022ab0947a9bce1b1209769b8d91/src/math/Color.js#L5

0b5vr (approved):
looks good!

0b5vr:
@all-contributor add rotu for code

PR #335, opened 2023-02-03
Cleanup deprecated code and Update documentation for BufferGeometry and BufferAttribute

rafaelsc:
### Why

BufferGeometry and BufferAttribute type need some cleanup to remove code that doesn't exist anymore on ThreeJS. And even marked was deprecated would cause runtime errors.  And some documentation spelling and naming needed improvements.

### What

- Remove the old code
- Improve documentation spelling and naming
- Update some return Types with the correct type according to the Documentation and Source code.

### Checklist

-   [x] Checked the target branch (current goes `master`, next goes `dev`)
-   [x] Added myself to contributors table
-   [x] Ready to be merged

PR #354, opened 2023-02-21
Update documentation for ThreeJS Core Classes

rafaelsc:
### Why

Updating the documentation will take some time and will need to be done in parts. This PR will update the Core.* (expect BufferGeometry and BufferAttribute, that is done in another PR)

### What

- Improve documentation for many `Core.*` classes 
  - BufferGeometry and BufferAttribute update in another PR - #335
- Remove some old code deprecated code.
- Update some return Types with the correct type according to the Documentation and Source code.
- Code Format

### Checklist

-   [ ] Added myself to contributors table. _Added on previous PR #335_
-   [x] Ready to be merged

### DIFF
--- a/types/three/src/core/Clock.d.ts
+++ b/types/three/src/core/Clock.d.ts
@@ -3,7 +3,7 @@
  * @remarks
  * This uses {@link https://developer.mozilla.org/en-US/docs/Web/API/Performance/now | performance.now} if it is available,
  * otherwise it reverts to the less accurate {@link https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date/now | Date.now}.
- * @see {@link https://threejs.org/docs/index.html#/api/en/core/Clock | Official Documentation}
+ * @see {@link https://threejs.org/docs/index.html#api/en/core/Clock | Official Documentation}
  * @see {@link https://github.com/mrdoob/three.js/blob/master/src/core/Clock.js | Source}
  */
 export class Clock {
--- a/types/three/src/core/Clock.d.ts
+++ b/types/three/src/core/Clock.d.ts
@@ -44,7 +44,7 @@ export class Clock {
     running: boolean;
 
     /**
-     * Starts clock
+     * Starts clock.
      * @remarks
      * Also sets the {@link startTime | .startTime} and {@link oldTime | .oldTime} to the current time,
      * sets {@link elapsedTime | .elapsedTime} to `0` and {@link running | .running} to `true`.
--- a/types/three/src/core/Clock.d.ts
+++ b/types/three/src/core/Clock.d.ts
@@ -9,24 +9,24 @@
 export class Clock {
     /**
      * Create a new instance of {@link THREE.Clock | Clock}
-     * @param autoStart - Whether to automatically start the clock when {@link Clock.getDelta | .getDelta}() is called for the first time. Default `true`
+     * @param autoStart - Whether to automatically start the clock when {@link getDelta | .getDelta()} is called for the first time. Default `true`
      */
     constructor(autoStart?: boolean);
 
     /**
-     * If set, starts the clock automatically when {@link Clock.getDelta | .getDelta}() is called for the first time.
+     * If set, starts the clock automatically when {@link getDelta | .getDelta()} is called for the first time.
      * @defaultValue `true`
      */
     autoStart: boolean;
 
     /**
-     * Holds the time at which the clock's {@link Clock.start | start} method was last called.
+     * Holds the time at which the clock's {@link start | .start()} method was last called.
      * @defaultValue `0`
      */
     startTime: number;
 
     /**
-     * Holds the time at which the clock's {@link Clock.start | start}, {@link Clock.getElapsedTime | .getElapsedTime}() or {@link Clock.getDelta | .getDelta}() methods were last called.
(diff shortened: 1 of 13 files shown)


0b5vr:
@rafaelsc I'm going to review this.

How did you do this by the way? You've done all of them by hand??

rafaelsc:
Thanks for taking some time to review @0b5vr .

I create an HTML parser for the Official Docs, that returns to me the formatted TSDocs. For now is 90% good enough.
image


At this moment I wanted to review the generated code, so I did some manual copy-paste. But my end goal is to learn the TS Compiler internals to make it possible to read and inject the Tsdocs automatically. But for that, I will need more time.

0b5vr:
That looks so promising! The solution already seems very convenient.
In this case, I will skim to see if there are any obvious errors.

types/three/src/core/Clock.d.ts
```diff
@@ -1,64 +1,72 @@
...
-     * This property keeps track whether the clock is running or not.
-     * @default false
+     * Whether the clock is running or not.
+     * @defaultValue `false`
      */
     running: boolean;
 
     /**
-     * Starts clock.
+     * Starts clock
```
0b5vr: nits

```suggestion
     * Starts clock.
```

types/three/src/core/EventDispatcher.d.ts
```diff
@@ -13,12 +13,31 @@ export type EventListener<E, T, U> = (event: E & { type: T } & { target: U }) =>
...
+ * @see {@link https://threejs.org/docs/index.html#api/en/core/EventDispatcher | Official Documentation}
+ * @see {@link https://github.com/mrdoob/three.js/blob/master/src/core/EventDispatcher.js | Source}
  */
 export class EventDispatcher<E extends BaseEvent = Event> {
     /**
-     * Creates eventDispatcher object. It needs to be call with '.call' to add the functionality to an object.
+     * Creates {@link THREE.EventDispatcher | EventDispatcher} object.
+     * @remarks
+     * It needs to be call with {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call | Function.prototype.call()}
+     * to add the functionality to an object.
```
0b5vr: Is this a remark that is still active? I feel like it's a stale description since we are using classes now

types/three/src/core/GLBufferAttribute.d.ts
```diff
@@ -1,27 +1,121 @@
 /**
- * see {@link https://github.com/mrdoob/three.js/blob/master/src/core/GLBufferAttribute.js|src/core/GLBufferAttribute.js}
+ * This buffer attribute class does not construct a VBO
```
0b5vr: nits

```suggestion
 * This buffer attribute class does not construct a VBO.
```

0b5vr:
Links look like this in VSCode.
For external URL links, each link is led with a `|`.

It seems TSDoc is yet to finalize the notation of `@link`, so it might be valid or not.
I think it's okay to be as is.

image

types/three/src/core/InterleavedBufferAttribute.d.ts
```diff
@@ -2,51 +2,191 @@ import { BufferAttribute } from './BufferAttribute';
...
+     * @param offset
+     * @param normalized Default `false`.
+     */
     constructor(interleavedBuffer: InterleavedBuffer, itemSize: number, offset: number, normalized?: boolean);
 
     /**
-     * @default ''
+     * The offset in the underlying array buffer where an item starts.
+     * @remarks Expects a `Integer`
+     * @defaultValue `''`
```
0b5vr: This seems to be wrong.

types/three/src/core/Object3D.d.ts
```diff
@@ -218,146 +255,165 @@ export class Object3D<E extends BaseEvent = Event> extends EventDispatcher<E> {
...
+     * @param z Expects a `Float`
+     */
     lookAt(x: number, y: number, z: number): void;
 
     /**
-     * Adds object as child of this object.
+     * Adds another {@link Object3D} as child of this {@link Object3D}.
+     * @remarks An arbitrary number of objects may be added
+     * @remarks Any current parent on an {@link object} passed in here will be removed, since an {@link Object3D} can have at most one parent.
+     * @see {@link attach}
```
0b5vr, line 406: very cool remark

rafaelsc:
@0b5vr What VSCode version and OS are you using? Here using the last version on Windows looks good.
image

types/three/src/core/InterleavedBufferAttribute.d.ts
rafaelsc: Good find. Fixed.

types/three/src/core/EventDispatcher.d.ts
rafaelsc: The Implementation Internally still uses .call. But I almost removed this doc from the old version and decided to keep it. But no problem I'm happy removing too. Fixed.

0b5vr: I see! I think it's now a deprecated use even internally so it's fine to remove from docs side

0b5vr:
> @0b5vr What VSCode version and OS are you using? Here using the last version on Windows looks good.

Ahh, the typescript version was 4.9.5 (VSCode) instead of 5.0.0-dev (Workspace). it looks fine on 5.0.0!

0b5vr (approved):
Looks good! Thank you for the fantastic work!

PR #356, opened 2023-02-24
Fix usage with "module": "nodenext"

### LINKED ISSUE #336: three/addons alias

Since r144 (https://github.com/mrdoob/three.js/pull/23406), three.js supports a `three/addons` alias that would currently redirect to `three/examples/jsm`.

For instance, the following resolve to the same file:

```js
import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js'
import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js'
```

### Suggested implementation:

TypeScript doesn't recognize the above behavior since the alias is currently handled through `package.exports` and types only exist for `examples/jsm` on disk (this might be something to check in TypeScript upstream). We can implement this by copying `examples/jsm` to `addons` at build-time if there's no easy way to alias as a library.

You can work-around this in user-land with:

```json
{
  "compilerOptions": {
    "paths": {
      "three/addons/*": ["./node_modules/three/examples/jsm/*"]
    }
  }
}
```

Methuselah96:
### Why

I was a little too hasty with https://github.com/three-types/three-ts-types/pull/352 and didn't test it with `"module": "nodenext"`. Turns out if you specify `"type": "module"` it expects all the imports to have extensions. DefinitelyTyped doesn't currently support this, but they are working on it. Until then, we should remove `"type": "module"`.

This also resolves https://github.com/three-types/three-ts-types/issues/336, since the `exports` field works when using `"module": "nodenext"`.

### What

Remove `"type": "module"` (reverting https://github.com/three-types/three-ts-types/pull/352). Also add `build/three.d.cts` to match `build/three.cjs`.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/OTHER_FILES.txt
+++ b/types/three/OTHER_FILES.txt
@@ -1,3 +1,4 @@
+build/three.d.cts
 build/three.d.ts
 build/three.module.d.ts
 examples/jsm/animation/AnimationClipCreator.d.ts
--- a/types/three/build/three.d.cts
+++ b/types/three/build/three.d.cts
@@ -1,0 +1,2 @@
+export * from '../src/Three';
+export as namespace THREE;
--- a/types/three/package.json
+++ b/types/three/package.json
@@ -1,6 +1,5 @@
 {
     "private": true,
-    "type": "module",
     "exports": {
         ".": {
             "import": "./build/three.module.js",


Methuselah96:
Also resolves https://github.com/three-types/three-ts-types/issues/336.

joshuaellis (approved):

PR #359, opened 2023-03-01
Allow Color to be set from an InterleavedBufferAttribute

Methuselah96:
### Why

A `Color` can also be set from an `InterleavedBufferAttribute` since [the code for that method](https://github.com/mrdoob/three.js/blob/master/src/math/Color.js#L574-L582) is just calling `getX()`/`getY()`/`getZ()`.

### What

Add `InterleavedBufferAttribute` as an allowed type for `Color.fromBufferAttribute`.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/math/Color.d.ts
+++ b/types/three/src/math/Color.d.ts
@@ -2,6 +2,7 @@ import { ColorSpace } from '../constants';
 import { ColorRepresentation } from '../utils';
 
 import { BufferAttribute } from './../core/BufferAttribute';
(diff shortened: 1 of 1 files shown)


0b5vr:
The implementation of `getX`, `getY`, and `getZ` on `InterleavedBufferAttribute`:

https://github.com/mrdoob/three.js/blob/47b28bc564b438bf2b80d6e5baf90235292fcbd7/src/core/InterleavedBufferAttribute.js#L129-L157

0b5vr (approved):
This should work indeed

PR #364, opened 2023-03-02
Re-export types from fflate

Methuselah96:
### Why

We should just depend on `fflate` and re-export the types from there instead of them duplicating them here.

### What

Re-export types from `fflate`.

### Checklist

-   [x] Ready to be merged


Methuselah96:
Waiting for https://github.com/microsoft/DefinitelyTyped-tools/pull/627 to get released.

types/three/package.json
```diff
@@ -10,5 +10,8 @@
         "./addons/*": "./examples/jsm/*",
         "./src/*": "./src/*",
         "./nodes": "./examples/jsm/nodes/Nodes.js"
+    },
+    "dependencies": {
+        "fflate": "~0.6.9"
```
joshuaellis: Does webxr not need to be here then? 🤔

Methuselah96: IIRC, any `@types/*` packages get added into the `dependencies` at publish-time by DefinitelyTyped, and therefore don't need to be specified here.

joshuaellis: Cool! thanks for explaining 😄

joshuaellis (approved):

PR #366, opened 2023-03-03
Fix OctreeHelper constructor

Methuselah96:
### Why

The second parameter of `OctreeHelper` is optional ([source](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/helpers/OctreeHelper.js#L10)) and is used that way ([source](https://github.com/mrdoob/three.js/blob/dev/examples/games_fps.html#L439)).

### What

Make second parameter of the `OctreeHelper` constructor optional.

### Checklist

-   [x] Ready to be merged


0b5vr (approved):
Looks good!

Issue #368, opened 2023-03-04
EventDispatcher Typing is very limited.


rafaelsc:
### Problem description:

The current `EventDispatcher` typing is very limited, 

- A unusual way to define strong type events, very different from what other TS Typing projects use.
- When using strong-type events, there are not easy to extend the definition to add more events.
- There are no strong-type events that ThreeJS currently uses.
- Type for `event.target` is set to _any_ currently.
- Classes that are EventDispatcher needs to allow dispatch and listening to any custom event, even if we are using strong-type events.

### Suggested solution:

- Update the typing using the same pattern that TS uses for DOM events.
- Add strong-type events to all known Threejs events.
- Use _interfaces_ to alow an easy way to extend the current strong-type events definitions.

Methuselah96:
Resolved by https://github.com/three-types/three-ts-types/pull/369.

PR #369, opened 2023-03-04
New strong-type EventDispatcher and Threejs strong-types events

rafaelsc:
BREAKING CHANGE

- Old strong-typing EventDispatcher is not compatible with this new typing. Requiring migration.

### Why

- This PR is to improve the EventDispatcher typing solving the #368 

### What

- Update EventDispatcher Typing 
- Update EventDispatcher Typing testing
- Add strong-types events to current known ThreeJS events

### Checklist

-   [x] Ready to be merged


0b5vr:
I'm not too expertised to see if it's conventionally a good solution.
I see that the motivation is valid at least.

joshuaellis:
So just in a "eli5" way, you're passing what you expect the target to be now? e.g. like event handlers in react?

rafaelsc:
#### ELI5: 
The Goal is Events in ThreeJs can be strongly typed and allow any other _external lib_ or user of the _three-types_ extends with his own custom strong events.


#### Non ELI5:
The goal is for `THREE.EventDispatcher` to have the same typing capabilities as any DOM HTML Element strong typing events from typescript `dom.generated.d.ts`. This link has a good explanation of how DOM String event works in Typescript, and how we can extend them. https://43081j.com/2020/11/typed-events-in-typescript

Looking at the Typing of [HTMLElement](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L6536
) the strong typed Events are mapped via [HTMLElementEventMap](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L6532) with a hierarchy of **EventMap** interfaces, and this Map's the `addEventListener` can expose the correct strong typed event.
This PR will add these same features to `THREE.EventDispatcher`.

Other changes are:
- Type-checking when dispatching known events.
- Add **EventMap** with strong-typed events for all current know events fired by ThreeJS.
- Allow dispatch and Listener of unknowing custom events. _(Current not possible with current typing)_.
- Remove `any` from the fired `ev.target` to the correct type from the Dispatcher.
- Add a `Disposable` interface to use in any Class that has the `dispose()` method.

joshuaellis (approved):
Nice explanation, lovely addition.

types/three/examples/jsm/controls/ArcballControls.d.ts
```diff
@@ -156,3 +162,5 @@ export class ArcballControls extends EventDispatcher {
 
     dispose(): void;
 }
+
+export {};
```
Methuselah96: What is the purpose of these `export` statements?

Methuselah96:
Looks good from a first glance, will take a closer look soon.

types/three/src/core/EventDispatcher.d.ts
```diff
@@ -1,15 +1,37 @@
...
- * Event object.
+ * The minimal expected contract of a fired Event that was dispatched by a {@link EventDispatcher<>}.
  */
-export interface Event extends BaseEvent {
-    target?: any;
-    [attachment: string]: any;
+export interface Event<TSource = unknown, TEventType extends string = string> {
+    readonly type: TEventType;
+    readonly target: TSource;
+}
```
Methuselah96: Wouldn't `TTarget` be more accurate than `TSource`?

types/three/src/core/EventDispatcher.d.ts
```diff
@@ -1,15 +1,37 @@
...
+
+// tslint:disable-next-line:interface-over-type-literal - Type Aliases to add better readability.
+export type EmptyEvent = {};
+
+type EventMap = Record<string, {}>;
+type EventKey<T extends EventMap> = string & keyof T;
+
+type EventReceiver<TSource, TEventType extends string, TEventData> = (
+    event: TEventData & Event<TSource, TEventType>,
+) => void;
```
Methuselah96: How about keeping this called `EventListener` and keeping the order of the generics the same as before to avoid the change in this type being a breaking change?

types/three/src/core/EventDispatcher.d.ts
```diff
@@ -1,15 +1,37 @@
...
+type EventTypeValidator<TEvent extends BaseEvent, TEventMap extends {}> = TEvent extends {
+    type: infer TEventType;
 }
-export type EventListener<E, T, U> = (event: E & { type: T } & { target: U }) => void;
+    ? TEventType extends EventKey<TEventMap>
+        ? { type: TEventType } & TEventMap[TEventType]
+        : TEventType extends string
+        ? TEvent
+        : never
+    : never;
```
Methuselah96: So this supports dispatching basically any event if the key does not exist in the event map. Are there common use-cases where this is necessary? Seems like it might be nice to have the ability to block any undeclared events.

types/three/test/core/core-eventDispatcher.ts
```diff
@@ -30,15 +37,33 @@ eveDisForTestEvent.addEventListener('foo', e => {
     e.bar;
 });
 // call addEventListener with an invalid event
-// @ts-expect-error
-eveDisForTestEvent.addEventListener('baz', () => {});
+// ts-expect-error
+eveDisForTestEvent.addEventListener('baz', e => {
+    e.type; // $ExpectType "baz"
+    e.target; // $ExpectType EventDispatcher<TestEvent>
+});
```
Methuselah96: It looks we're no longer producing an error in this case since `@ts-expect-error` was changed to `ts-expect-error`. Is that intentional?

Methuselah96 (changes requested):
Left some initial thoughts.

types/three/examples/jsm/controls/ArcballControls.d.ts
rafaelsc: Because the `ArcballControlsEventMap` is not exported.
DefinitelyTyped rules complain that there is a non-exported type. 
This is the way to tell the DefinitelyTyped rule that you want not to export something by default.  
_This is used in many places to allow not export the EventMap_

types/three/src/core/EventDispatcher.d.ts
rafaelsc: `EventDispatcher` allow dispatch and listen for any event.
In my opinion, Typescript types should not block or reduce the features/capabilities of the Javascript code, only add helpful type information to reduce the errors caused by not using the correct typing.
Blocking unknown event types would reduce the features/capabilities of the original `EventDispatcher` code.

I don't know how common is for people to use custom events on `THREE.Object3D` or custom `EventDispatcher `. but I imagine that some are. And I think EventDispatcher typing should allow them to decide if they want the custom event to be strong-typed or not.

I can imagine that I would prefer to create a simple quick and dirty custom `Object3D` or other `EventDispatcher` sub-type and test each feature without wanting to have to do the bureaucracy of creating custom strong-type events for a test.

types/three/src/core/EventDispatcher.d.ts
rafaelsc: Update with your suggestion.

types/three/src/core/EventDispatcher.d.ts
rafaelsc: Update with your suggestion.

types/three/test/core/core-eventDispatcher.ts
rafaelsc: Sorry, that was an error. Tests are now updated and with better comments.

types/three/src/core/EventDispatcher.d.ts
Methuselah96: I think it's reasonable to want to block unknown event types if the user knows exactly what set of event types they want to allow.

It seems like we could support both approaches. For the places where we want the event type to be open, the event map could include an index signature. For the places where we want the event type to be closed, the event map could not include an index signature.

My problem with the current approach is that it makes it impossible to block unknown events, whereas relying on whether there's an index signature or not would allow for either approach.

Thoughts?

rafaelsc:
Sorry, I finally fixed the conflict.

Methuselah96:
@rafaelsc I'm interested in completing this, but I don't want to step on your toes. Let me know if you plan to get back this, otherwise I will probably take this on.

types/three/src/core/EventDispatcher.d.ts
Methuselah96: I suppose there's no way to make a closed hierarchy of events in the way DOM events are typed either. Let me reconsider this.

Methuselah96:
To follow-up on the discussion regarding [whether to allow dispatching unknown events](https://github.com/three-types/three-ts-types/pull/369#discussion_r1147501968).

After trying a few approaches, taking a more strict approach to events and allowing the user to loosen the typings seems like the best option.

### Rationale

The main reason for taking this approach is that it is the only safe approach. Consider a simplified `XRHandSpace` definition:

```ts
interface Object3DEventMap {
    added: {};
    removed: {};
}

interface WebXRSpaceEventMap extends Object3DEventMap {
    select: { data: XRInputSource };
}

class XRHandSpace extends Group<WebXRSpaceEventMap> {}
```

If we allowed dispatching unknown events, it would allow dispatching a bad `select` event if the `XRHandSpace` is considered as an `Object3D` like this:

```ts
const handSpace = new THREE.XRHandSpace();

function dispatchSelectEvent(object: THREE.Object3D) {
    object.dispatchEvent({ type: 'select' });
}

dispatchSelectEvent(handSpace);
```

This breaks how inheritance and types should work and it seems best to at least provide the option of avoiding this issue.

### Dispatching unknown events

Because of that, we only allow dispatching defined events. There are a few ways to add events to the defined event map.

1. If you're just creating an `Object3D`, then you can define the event map in the generic:

```ts
const allowSpecificEvent = new THREE.Object3D<THREE.Object3DEventMap & { test: { data: number } }>();
allowSpecificEvent.dispatchEvent({ type: 'test', data: 5 });

const allowAnyEvent = new THREE.Object3D<THREE.Object3DEventMap & Record<string, { [key: string]: unknown }>>();
allowAnyEvent.dispatchEvent({ type: 'any', data: 5 });
```

2. If you're creating a new class, you can define it in the class definition:

```ts
class AllowSpecificEvent extends Object3D<THREE.Object3DEventMap & { test: { data: number } }> {}
const allowSpecificEventClass = new AllowSpecificEvent();
allowSpecificEventClass.dispatchEvent({ type: 'test', data: 5 });

class AllowAnyEvent extends Object3D<THREE.Object3DEventMap & Record<string, { [key: string]: unknown }>> {}
const allowAnyEventClass = new AllowAnyEvent();
allowAnyEventClass.dispatchEvent({ type: 'any', data: 5 });
```

3. If you just want to allow to dispatch any event for any `Object3D` you could augment the `Object3DEventMap` interface:

```ts
declare module 'three' {
    interface Object3DEventMap {
        [key: string]: { [key: string]: unknown };
    }
}

const allowAnyEvent = new THREE.Object3D();
allowAnyEvent.dispatchEvent({ type: 'any', data: 5 });
```

### Conclusion

I'm hoping this change provides more good than harm, please let me know if this ends up causing a lot of churn and we can take a look at what we can do to make it better.

Methuselah96 (approved):

Issue #357, opened 2023-02-27
r150


0b5vr:
## Issue Tracker

This is used to track potential issues we need to investigate surrounding the `r150` release.

### Core

- [x] https://github.com/mrdoob/three.js/pull/24940
    - `ColorManagement.legacyMode = false` -> `ColorManagement.enabled = true`
    - #374
- [x] https://github.com/mrdoob/three.js/pull/24975
    - `renderer.physicallyCorrectLights = true` -> `renderer.useLegacyLights = false`
    - #374
- [x] https://github.com/mrdoob/three.js/pull/25418
    - Add `stopListenToKeyEvents()` to `OrbitControls`
    - #376
- [x] https://github.com/mrdoob/three.js/pull/25520
    - Add `DisplayP3ColorSpace` as a constant
    - #377
- [ ] https://github.com/mrdoob/three.js/pull/25549
    - We don't have a type definition for `WebGLMorphTargets`. Would we add this?

### Docs

- [ ] Docs changes

### Addons

- [ ] Node material related changes
- [ ] WebGPU related changes

Last checked: https://github.com/mrdoob/three.js/releases/tag/r150

Methuselah96:
[Released](https://github.com/three-types/three-ts-types/releases/tag/r150)

PR #372, opened 2023-03-13
fix(ColorManagement): enabled => !legacyMode

CodyJasonBennett:
### Why

`ColorManagement.legacyMode` was renamed to `ColorManagement.enabled` which inverts the flag (https://github.com/mrdoob/three.js/pull/24940).

### What

Three.js deprecation policy gives this 10 releases until it's removed, with a warning in-between, but I've removed it here.

### Checklist

-   [x] Ready to be merged


Methuselah96:
Thanks, for the PR, I think this should be based against `dev` since it's for [the r150 release](https://github.com/three-types/three-ts-types/issues/357).

CodyJasonBennett:
Sorry, tried doing this on mobile.

PR #379, opened 2023-03-13
RaycasterParameters should also support Line2

promontis:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64610

### DIFF
--- a/types/three/src/core/Raycaster.d.ts
+++ b/types/three/src/core/Raycaster.d.ts
@@ -35,6 +35,7 @@ export interface Intersection<TIntersected extends Object3D = Object3D> {
 export interface RaycasterParameters {
     Mesh?: any;
     Line?: { threshold: number } | undefined;
+    Line2?: { threshold: number } | undefined;
     LOD?: any;
     Points?: { threshold: number } | undefined;
     Sprite?: any;


0b5vr:
I assume that users might manually set `raycaster.params.Line2 = { threshold: 1 }` if they wish to utilize the threshold.

It's weird that the default `params` does not have the `Line2` though...

https://github.com/mrdoob/three.js/blob/47b28bc564b438bf2b80d6e5baf90235292fcbd7/src/core/Raycaster.js#L16-L22

Mid% confidence but I understood and I will approve this.

0b5vr (approved):

promontis:
@0b5vr yeah, it is kinda strange that the default `params` didn't include it yet. I originally also was setting `Line`, thinking that it would use that threshold, since a line is a line. But the code actually uses Line2, which took me some time before browsing ThreeJS's code, seeing how it is used. Can this now also be merged? Like, who hits the merge button?

0b5vr:
@Methuselah96 Does it look good? Plus I was not sure I can merge this into `master` right now because it seems that you're working on the r150 release.

Methuselah96:
One of the problems is that `Line2` is not in core, and in general it doesn't seem like a good idea for something in the core to depend on an example. It seems that `Line2` is not special in that anyone could override `Object3D#raycast` and pull anything they like off of `raycater.params`.

That being said, this seems like most straight-forward fix. We could make `RaycasterParameters` extend `Record<string, unknown>` in order to allow any parameter to be set, but then we lose auto-complete for `Line2`. Let's just go with this for now.

PR #387, opened 2023-03-27
Revert MOUSE and TOUCH back to enums

Methuselah96:
### Why

https://github.com/three-types/three-ts-types/pull/349 made all of the constants non-enums. While this makes sense for most of the constants, the `MOUSE` and `TOUCH` were the only constants that were indistinguishable from an enum. In an effort to avoid making a breaking change, I think it makes sense to convert them back to enums until it causes issues.

### What

Revert MOUSE and TOUCH back to enums

### Checklist

-   [x] Ready to be merged


Methuselah96:
I'm going to go ahead and merge this so we can get a patch out quickly before this causes issues for users.

PR #393, opened 2023-04-02
Update documentation for ThreeJS extra Classes, with some Typing improvements

rafaelsc:
### Why

Some Typing information is missing and could be improved.

### What

- Improve documentation for all extra.* classes/files.
- Add Missing Definition Type
- Update some Types 
- Fix some wrong  Types 
- Code Format

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/extras/core/Interpolations.d.ts
+++ b/types/three/src/extras/core/Interpolations.d.ts
@@ -5,6 +5,8 @@
  * @param p1 Expects a `Float`
  * @param p2 Expects a `Float`
  * @param p3 P0, p1, p2, the points defining the spline curve. Expects a `Float`
+ * @see {@link https://threejs.org/docs/index.html#api/en/extras/core/Interpolations | Official Documentation}
+ * @see {@link https://github.com/mrdoob/three.js/blob/master/src/extras/core/Interpolations.js | Source}
  */
 declare function CatmullRom(t: number, p0: number, p1: number, p2: number, p3: number): number;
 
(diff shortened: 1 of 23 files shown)


types/three/src/extras/core/Interpolations.d.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param p0 Expects a `Float`
+ * @param p1 Expects a `Float`
+ * @param p2 Expects a `Float`
+ * @param p3 P0, p1, p2, the starting, control(twice) and end points defining the curve. Expects a `Float`
+ * @see {@link https://threejs.org/docs/index.html#api/en/extras/core/Interpolations | Official Documentation}
+ * @see {@link https://github.com/mrdoob/three.js/blob/master/src/extras/core/Interpolations.js | Source}
+ */
+declare function CubicBezier(t: number, p0: number, p1: number, p2: number, p3: number): number;
+
+export { CatmullRom, QuadraticBezier, CubicBezier };
```
Methuselah96, line 36: What's going on with this file? I don't see it in three.js, and in general I like to keep things as close to the three.js core as possible.

Methuselah96 (changes requested):

Methuselah96 (approved):

types/three/src/extras/core/Interpolations.d.ts
rafaelsc: I can see this file in three.js. `https://github.com/mrdoob/three.js/blob/master/src/extras/core/Interpolations.js`
https://threejs.org/docs/index.html?q=inter#api/en/extras/core/Interpolations

So I added this missing file definition on three.js types.

PR #394, opened 2023-04-02
Update documentation for ThreeJS helpers Classes, with some Typing improvements

rafaelsc:
### Why

Some Typing information is missing and could be improved.

### What

- Improve documentation for all helpers.* classes/files.
- Update some Types 
- Fix some wrong Types 
- Code Format

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/helpers/ArrowHelper.d.ts
+++ b/types/three/src/helpers/ArrowHelper.d.ts
@@ -4,16 +4,32 @@ import { Mesh } from './../objects/Mesh';
 import { Object3D } from './../core/Object3D';
 import { ColorRepresentation } from '../utils';
 
-// Extras / Helpers /////////////////////////////////////////////////////////////////////
-
+/**
+ * An 3D arrow object for visualizing directions.
+ * @example
+ * ```typescript
+ * const dir = new THREE.Vector3(1, 2, 0);
+ * //normalize the direction vector (convert to vector of length 1)
+ * dir.normalize();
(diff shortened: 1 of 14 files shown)


types/three/src/helpers/ArrowHelper.d.ts
```diff
@@ -25,9 +41,12 @@ export class ArrowHelper extends Object3D {
...
 
     /**
-     * @default 'ArrowHelper'
+     * A Read-only _string_ to check if `this` object type.
+     * @remarks Sub-classes will update this value.
+     * @override
+     * @defaultValue `ArrowHelper`
      */
-    type: string;
+    override readonly type: string | 'ArrowHelper';
```
Methuselah96, line 49: I'm curious why this can't be typed as just `'ArrowHelper'` here and in the rest of the files.

Methuselah96 (changes requested):

Methuselah96 (approved):

types/three/src/helpers/ArrowHelper.d.ts
rafaelsc: This is because of the base `Object3D` type. And the need for `type` de overridden in many sub-classes. You can only override a string type, not a string const. 
At the end Typescript will only consider this be a string, ignoring the `ArrowHelper` string const. But I decided put it there just for documentation and to help the developers when looking at the type definition to see what the default value he can expect.

PR #395, opened 2023-04-02
fix: update map controls location

oetherington:
### Why

The `MapControls` and `OrbitControls` classes used to be located in the same file, but https://github.com/mrdoob/three.js/pull/25608 moved `MapControls` out onto its own.

### What

This PR mirrors the change by moving the type definition for `MapControls` into a new `MapControls.d.ts` file, and adds in the updated doc comments.

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #399, opened 2023-04-10
fix: allow any string in ColorRepresentation

Shakhriddin:
Is it possible to add a string type to ColorRepresentation, now I want to update the version of three.js and I have an error on creating Color class, because the data comes as a string and the string is not specified here. Otherwise, it will be necessary to rewrite "as ColorRepresentation" everywhere

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/Three.d.ts
+++ b/types/three/src/Three.d.ts
@@ -3,7 +3,6 @@
  */
 export * from './constants';
 export * from './Three.Legacy';
-export * from './utils';
 /**
  * Animation
  */
--- a/types/three/src/helpers/ArrowHelper.d.ts
+++ b/types/three/src/helpers/ArrowHelper.d.ts
@@ -2,7 +2,7 @@ import { Vector3 } from './../math/Vector3';
 import { Line } from './../objects/Line';
 import { Mesh } from './../objects/Mesh';
 import { Object3D } from './../core/Object3D';
-import { ColorRepresentation } from '../utils';
+import { ColorRepresentation } from '../math/Color';
 
 // Extras / Helpers /////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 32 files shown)


Methuselah96:
@Shakhriddin Thanks for the PR! Just so that we can make a more informed decision, when you say "the data comes as a string", who defines the data, and are you not able to update the data type to more specific (i.e., update the color string to be  `ColorKeyword | ColorModelString | HexColorString` instead of `string`)?

Shakhriddin:
Thanks for the response, I understand that it's kind of useful to specify the type, but the data comes as a string from the backend and we have a whole library that has legacy code and legacy(old) libraries. I can fix the our legacy code, but I can't fix the libraries, they are very old. Will it be possible to add the string type. Thanks for the answer @Methuselah96

Methuselah96:
Thanks, I understand the situation. @joshuaellis @0b5vr Do you all have thoughts? I'm not really sure what the best solution is.

Methuselah96:
My current inclination is that we should revert https://github.com/three-types/three-ts-types/pull/350 and keep the type as `Color | string | number`. It sounds nice in theory to type-check the strings, but I think it ends up making the types too strict especially when trying to integrate with other systems that only know that it's a `string`.

Shakhriddin:
Thank you for understanding, I will wait for the next patch. @Methuselah96

joshuaellis:
> My current inclination is that we should revert #350 and keep the type as `Color | string | number`. It sounds nice in theory to type-check the strings, but I think it ends up making the types too strict especially when trying to integrate with other systems that only know that it's a `string`.

im fine with this change, I agree it's too strict and it's not exactly bullet proof i could write `#ajnasdjn2jno` which isn't a color.

Methuselah96 (approved):

Methuselah96:
@rotu FYI, we had to essentially revert https://github.com/three-types/three-ts-types/pull/350 since it can be a little bit too hard to integrate with other systems. Sorry about that.

Shakhriddin:
Thanks!

PR #401, opened 2023-04-11
Add Skeleton.toJSON, Skeleton.fromJSON

Hoodgail:
Addding a  Skeleton.toJSON since it was missing in the types.

https://github.com/mrdoob/three.js/blob/0fbae6f682f6e13dd9eb8acde02e4f50c0b73935/src/objects/Skeleton.js#L248

### DIFF
--- a/types/three/src/objects/Skeleton.d.ts
+++ b/types/three/src/objects/Skeleton.d.ts
@@ -114,4 +114,6 @@ export class Skeleton {
     dispose(): void;
     
     toJSON(): any;
+    
+    fromJson(json: any, bones: Record<string, Bone>): void;
 }
--- a/types/three/src/objects/Skeleton.d.ts
+++ b/types/three/src/objects/Skeleton.d.ts
@@ -113,7 +113,7 @@ export class Skeleton {
      */
     dispose(): void;
     
-    toJSON(): any;
+    toJSON(): unknown;
     
-    fromJSON(json: any, bones: Record<string, Bone>): void;
+    fromJSON(json: unknown, bones: Record<string, Bone>): void;
 }
--- a/types/three/src/objects/Skeleton.d.ts
+++ b/types/three/src/objects/Skeleton.d.ts
@@ -112,8 +112,8 @@ export class Skeleton {
      * Call this method whenever this instance is no longer used in your app.
      */
     dispose(): void;
-    
+
     toJSON(): unknown;
-    
+
     fromJSON(json: unknown, bones: Record<string, Bone>): void;
 }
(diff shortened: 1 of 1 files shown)


types/three/src/objects/Skeleton.d.ts
```diff
@@ -112,4 +112,8 @@ export class Skeleton {
      * Call this method whenever this instance is no longer used in your app.
      */
     dispose(): void;
+    
+    toJSON(): any;
+    
+    fromJSON(json: any, bones: Record<string, Bone>): void;
```
Methuselah96: Could we type the return from `toJSON` and the first parameter of `fromJSON`?

Methuselah96 (changes requested):

types/three/src/objects/Skeleton.d.ts
Hoodgail: Why? I think i'd be best to do type return for toJSON once we do type return for every other classes that have toJSON

like 
Object3D.toJSON
BufferGeometry.toJSON

Methuselah96: To avoid the use of `any`, since it can cause surprising results if a user doesn't realize it's `any`. If we don't want to type it fully, then let's just use `unknown` instead of `any` to avoid the appearance of it being typed.

Methuselah96: As for whether to wait until we do it for all the other classes, I'm more inclined to think doing the right thing here is more important than consistency across classes. If we wait until we can do it for all the classes at the same time, it's less likely to happen in a timely manner in my experience. Better to just get it right where we can and fix the other classes later.

Hoodgail: > As for whether to wait until we do it for all the other classes

image

image

😏 😏 😏 😏 😏

Hoodgail: If i where to push this, how would i go about it?? have a dedicated ts files for all of these jsons types and use it something like this
`Sterilized.Object3D`

Or have each in their corresponding files like

export interface `Object3DJSON`

idk your choice

Methuselah96: Ha, nice work! My preference is to keep them in the corresponding files with a name like `Object3DJSON` to keep it as close to the source as possible.

Hoodgail: > Ha, nice work! My preference is to keep them in the corresponding files with a name like `Object3DJSON` to keep it as close to the source as possible.

Alr, i think you should accept this pull request as is, then i will add the JSON types for everything including Skeleton in a new pull request, would make things easier.

I'll change it to unkown for now instead of any like you said

Methuselah96: Thanks.

Methuselah96 (approved):

PR #405, opened 2023-04-19
Fix types in ViewHelper

k2d222:
### Why

Incorrect types in the `ViewHelper` example.

### What

Fixed types in `examples/jsm/helpers/ViewHelper`:
* camera is a Camera, not an Octree
* MouseEvent is more generic than PointerEvent (only clientX and clientY are used).

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks. looks good! Rerunning the build, not sure why it was failing.

PR #420, opened 2023-04-26
fix(HemisphereLight): correct position modifier

demike:
the position member has to be readonly

### Why

because setting the position on HemisphereLight throws an error about readonly property

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks for the PR!

PR #426, opened 2023-04-26
Three json

Hoodgail:
- [x] Object3D
  - [x] Scene
  - [x] Mesh
    - [x] InstancedMesh
  - [x] Line
    - [x] LineLoop
    - [x] LineSegments
  - [x] Skeleton
  - [x] Camera 
    - [x] PerspectiveCamera
    - [x] OrthographicCamera
    - [x] CubeCamera
    - [x] ArrayCamera
  - [x] Group
  - [x] Bone
  - [x] Sprite
  - [x] Points
  - [x] LOD
  - [ ] Light
    - [x] HemisphereLight
    - [ ] SpotLight
    - [x] PointLight
    - [x] DirectionalLight
    - [x] AmbientLight
    - [ ] RectAreaLight
    - [ ] LightProbe 
  - [ ] Audio
    - [ ] PositionalAudio 
  - [ ] AudioListener

- [ ] FogBase
  - [ ] Fog
  - [ ] FogExp2

- [x] Math `(Core)`
  - [x] Euler
  - [x] Quaternion
  - [x] AnimationClip
  - [x] KeyframeTrack 
  - [x] Matrix3
  - [x] Matrix4
  - [x] Vector2
  - [x] Curve
    - [x] CurvePath
      - [x] Path

- [x] Texture
  - [ ] VideoTexture
  - [ ] DataTexture
  - [ ] CompressedTexture
  - [ ] CubeTexture
  - [ ] CanvasTexture
  - [ ] DepthTexture

- [x] Source 
- [x] BufferGeometry
- [x] BufferAttribute

- [x] KeyframeTrack
  - [x] BooleanKeyframeTrack
  - [x] ColorKeyframeTrack
  - [x] NumberKeyframeTrack
  - [x] QuaternionKeyframeTrack
  - [x] StringKeyframeTrack
  - [x] VectorKeyframeTrack

**Finalization**
- [ ] Implement the respective json interfaces to their toJSON functions.
- [ ] Revise


Hoodgail:
@Methuselah96 What do you think of the progress so far?
let me know if how i'm doing it is good before i continue

types/three/src/Meta.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface Meta<Type extends string, Generator extends string, Version extends number = 4.5> {
+    version: Version;
+    type: Type;
+    generator: Generator;
+}
```
Methuselah96: I would really like to avoid creating a file that doesn't exist in three.js. Looks likes this is used in 5 places. Could we either just in-line it in the places it's used (and get rid of the interface altogether) or move it into a file that already exists like `Object3D.d.ts`?

types/three/src/animation/KeyframeTrack.d.ts
```diff
@@ -4,6 +4,18 @@ import { CubicInterpolant } from './../math/interpolants/CubicInterpolant';
 import { Interpolant } from '../math/Interpolant';
 import { InterpolationModes } from '../constants';
 
+export interface KeyframeTrackJSON<Type extends string = 'KeyframeTrack'> {
+    type: Type;
```
Methuselah96: Can you help me understand how the `type` property works and why it's generic in all of them?

Methuselah96 (commented):
Looking good overall, thanks for working on this.

types/three/src/animation/KeyframeTrack.d.ts
Hoodgail: > Can you help me understand how the type property works and why it's generic in all of them?

Well first, i added it because i thought the type and generator in the Meta changed based on the type. but i was wrong and just kept it generic on all of them.

But it might be good for extended objects, like bone which is just Object3D with the .type as "Bone" 
I was thinking it would be better than overriding the type property, like Group

> I would really like to avoid creating a file that doesn't exist in three.js. Looks likes this is used in 5 places. Could we either just in-line it in the places it's used (and get rid of the interface altogether) or move it into a file that already exists like Object3D.d.ts?

Yeah i was thinking the same thing, i just didnt know what to do, i'll move it to `Object3D.d.ts`

I gave the meta its own interface just incase it changes, we woulnt have to change it in everything.
Since its pretty much the same, keeps from stuff being repetitive

Hoodgail:
@kourser Reminder to finish this.

Hoodgail:
*Note to self*
https://github.com/three-types/three-ts-types/blob/8109927fa28df9359537ea3cdc1b4c777b905a12/types/three/src/core/Object3D.d.ts#L646

expected output format:
```js
{
    geometries: geojson[]
    materials: materialjson[]
    textures: texturejson[]
    images: imagejson[]
    object: Object3DJSON
}
```

Methuselah96 (approved):
@Hoodgail Thanks for your work on this! I went ahead and did some cleanup of some of the changes, hopefully while preserving the intent of your changes. Let me know if I messed anything up and feel free to keep building on this with more PRs, I just figured this would be a good place to get some of it merged.

Hoodgail:
Thank you

mattrossman:
@Hoodgail  I traced this PR as the source of the `QuaternionTuple` change here

https://github.com/three-types/three-ts-types/blame/a6fadb5785285a73ccf9f9dbc8b81186c7ebfa42/types/three/src/math/Quaternion.d.ts#L14

Since `fromArray` is typed as
```ts
fromArray(array: QuaternionTuple, offset?: number): this;
```
it produces an error when provided an array that doesn't strictly have 4 elements, which is expected when passing an `offset`. This makes it impractical to use in common scenarios such as when the array is a `Float32Array` (e.g. packed buffer of multiple quaternions) or otherwise contains >4 elements.

CleanShot 2024-09-17 at 18 03 09@2x

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wCoAkBKBRPOAMyghDgHIYALKAUzooCgmBjCAO1XgEcBXZPAC8cDnQDumXAQB0ARQEw6UDsE4AKAJQt2XeDACeYOgBMAglCjIDcEWMkAxADYRBAZgBMFqwfUAObV1uOAgiIlQ6YTgAFhZ+QRkSMm9rdUNjc0trABoQsIiYTSA

Is there a reason `array` needs to be typed this way?

Hoodgail:
@mattrossman It looks like this was an oversight on my part. I had completely forgot about the offset argument

It should follow the same way its written in Vector3 at https://github.com/three-types/three-ts-types/blob/a6fadb5785285a73ccf9f9dbc8b81186c7ebfa42/types/three/src/math/Vector3.d.ts#L269
{D400AF8F-C00C-4E49-B28E-30EA41E23447}

Will you make a pull request? or should i? My apologies !

mattrossman:
@Hoodgail I've submitted a PR https://github.com/three-types/three-ts-types/pull/1230, thank you for the suggested fix!

PR #429, opened 2023-04-27
TSL changes

Methuselah96:
### Why

Make type changes corresponding to https://github.com/mrdoob/three.js/pull/25683 and https://github.com/mrdoob/three.js/pull/25685.

### What

Many updates.

### Checklist

-   [x] Ready to be merged


LeviPesin:
Do you consider porting https://github.com/mrdoob/three.js/pull/25498? It's a quite a large PR with quite a many breaking changes.

Methuselah96:
It is stated in [the README](https://github.com/three-types/three-ts-types#priorities-and-goals):

> It is not a goal for the `examples/jsm` directory to have complete declaration files for every examples module due to the number of example modules and how quickly they are modified. If you would like to use an examples module that is missing types then create a PR to add the types or create an issue to request that we add the missing types for that module.

Unfortunately I (we) don't have the bandwidth to keep pace with the changes made to the `nodes` directory. The most we have time for is removing types that are no longer accurate. We would love it if someone was able to contribute those changes though, and would be glad to review and merge them.

Methuselah96:
My priority at the moment is releasing [r152 type changes](https://github.com/three-types/three-ts-types/issues/433), but I may took a look at that PR (and the nodes types in general) after I'm done with getting r152 released.

Methuselah96:
@LeviPesin FYI, I finally got around to this in https://github.com/three-types/three-ts-types/pull/720. Still a few nodes missing, but at least got the `ShaderNodeElements` refactor out of the way.

PR #451, opened 2023-05-03
fix: wrong typing of SpotLight

demike:
position should be readonly

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #458, opened 2023-05-10
Define GridHelper Material generic

Methuselah96:
### Why

[`GridHelper` uses a `BufferGeometry` and `LineBasicMaterial`](https://github.com/mrdoob/three.js/blob/master/src/helpers/GridHelper.js#L34C24-L40). Narrowing the generic makes it easier to make updates to the material properties without having to consider whether the material is an array.

### What

Define geometry and material generics for `GridHelper`.

### Checklist

-   [x] Ready to be merged


0b5vr (approved):
looks good

PR #464, opened 2023-05-17
Add missing `pmxAnimation` parameter to `MMDAnimationHelperParameter`

noname0310:
### Why

see:
https://github.com/mrdoob/three.js/blob/dev/examples/jsm/animation/MMDAnimationHelper.js#L47

### What

add missing property `pmxAnimation`

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #466, opened 2023-05-25
Add missing functions and properties in several classes

repalash:
Changes in: OrbitControls, GLTFExporter, GLTFWriter, ObjectLoader, Material, Matrix3, Matrix4, WebGLMultipleRenderTargets

### Why

Add missing properties and functions.

### What

- Extend OrbitControls from EventDispatcher like in three.js
- Put GLTFWriter in GLTFExporter.Utils like in three.js
- Add functions to GLTFWriter
- Add `name` property in GLTFLoaderPlugin
- Add missing parse functions to GLTFLoader and fix return types of others.
- Add onBeforeRender in Material
- Add readonly props isMatrix3 and isMatrix4
- Add samples to MultisampleRenderTargets

-   [x] Ready to be merged

### DIFF
--- a/types/three/examples/jsm/exporters/GLTFExporter.d.ts
+++ b/types/three/examples/jsm/exporters/GLTFExporter.d.ts
@@ -1,4 +1,4 @@
-import {Object3D, AnimationClip, Texture, Material, Mesh, BufferGeometry, PixelFormat} from '../../../src/Three';
+import { Object3D, AnimationClip, Texture, Material, Mesh, BufferGeometry, PixelFormat } from '../../../src/Three';
 
 export interface GLTFExporterOptions {
     /**
@@ -89,7 +89,7 @@ export class GLTFExporter {
 
     static Utils: {
         GLTFWriter: typeof GLTFWriter;
-    }
+    };
 }
 
 declare class GLTFWriter {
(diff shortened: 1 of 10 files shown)


types/three/examples/jsm/exporters/GLTFExporter.d.ts
```diff
@@ -86,9 +86,13 @@ export class GLTFExporter {
         input: Object3D | Object3D[],
         options?: GLTFExporterOptions,
     ): Promise<ArrayBuffer | { [key: string]: any }>;
+
+    static Utils: {
+        GLTFWriter: typeof GLTFWriter;
+    };
```
Methuselah96, line 92: I see `Utils` being set on `GLTFExporter` but I'm having trouble seeing where `GLTFWriter` is declared on it. Can you show where that happens?

Methuselah96 (changes requested):
Thanks for the PR! I did not have time to review the whole thing, if there are parts you would like to get merged in sooner you could consider splitting this into smaller PRs since some of these changes are easier to verify than others.

repalash:
Hi @Methuselah96, No hurry, please take your time.

types/three/examples/jsm/exporters/GLTFExporter.d.ts
repalash: Oh, my bad, this was a change I was making a PR for three.js, forgot that its not there.
I will remove this change.

PR #467, opened 2023-05-25
Changes in Pass and RenderPass

repalash:
Changes in Pass and RenderPass

### What
Allow `WebGLMultipleRenderTargets` in `Pass.render` for read and write buffers. This is required because `WebGLMultipleRenderTargets` does not inherit from WebGLRenderTarget in this types repo(but it does in three.js)

Allow `readBuffer`, `deltaTime`, `maskActive` to be optional in Pass. 

Add render function in RenderPass, which is slightly different from Pass.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/examples/jsm/postprocessing/Pass.d.ts
+++ b/types/three/examples/jsm/postprocessing/Pass.d.ts
@@ -1,4 +1,4 @@
-import { Material, WebGLRenderer, WebGLRenderTarget } from '../../../src/Three';
+import {Material, WebGLMultipleRenderTargets, WebGLRenderer, WebGLRenderTarget} from '../../../src/Three';
 
 export class Pass {
     constructor();
@@ -12,10 +12,10 @@ export class Pass {
     setSize(width: number, height: number): void;
     render(
         renderer: WebGLRenderer,
-        writeBuffer: WebGLRenderTarget,
-        readBuffer: WebGLRenderTarget,
-        deltaTime: number,
-        maskActive: boolean,
+        writeBuffer: WebGLMultipleRenderTargets | WebGLRenderTarget | null,
+        readBuffer?: WebGLMultipleRenderTargets | WebGLRenderTarget,
+        deltaTime?: number,
+        maskActive?: boolean,
     ): void;
 
     dispose(): void;
--- a/types/three/examples/jsm/postprocessing/Pass.d.ts
+++ b/types/three/examples/jsm/postprocessing/Pass.d.ts
@@ -1,4 +1,4 @@
-import {Material, WebGLMultipleRenderTargets, WebGLRenderer, WebGLRenderTarget} from '../../../src/Three';
+import { Material, WebGLMultipleRenderTargets, WebGLRenderer, WebGLRenderTarget } from '../../../src/Three';
 
 export class Pass {
     constructor();
--- a/types/three/examples/jsm/postprocessing/RenderPass.d.ts
+++ b/types/three/examples/jsm/postprocessing/RenderPass.d.ts
@@ -1,4 +1,12 @@
-import {Scene, Camera, Material, Color, WebGLMultipleRenderTargets, WebGLRenderTarget, WebGLRenderer} from '../../../src/Three';
+import {
+    Scene,
+    Camera,
+    Material,
+    Color,
+    WebGLMultipleRenderTargets,
+    WebGLRenderTarget,
+    WebGLRenderer,
+} from '../../../src/Three';
 import { Pass, FullScreenQuad } from './Pass';
 
 export class RenderPass extends Pass {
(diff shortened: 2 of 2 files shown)


types/three/examples/jsm/postprocessing/RenderPass.d.ts
```diff
@@ -1,13 +1,28 @@
...
 
 export class RenderPass extends Pass {
-    constructor(scene: Scene, camera: Camera, overrideMaterial?: Material, clearColor?: Color, clearAlpha?: number);
-    scene: Scene;
-    camera: Camera;
-    overrideMaterial: Material;
-    clearColor: Color;
+    constructor(scene?: Scene, camera?: Camera, overrideMaterial?: Material, clearColor?: Color, clearAlpha?: number);
+    scene?: Scene;
+    camera?: Camera;
```
Methuselah96, line 15: Can you provide an example or show in the code where `scene` or `camera` can be optional/undefined?

types/three/examples/jsm/postprocessing/RenderPass.d.ts
```diff
@@ -1,13 +1,28 @@
...
+    overrideMaterial?: Material;
+    clearColor?: Color;
     clearAlpha: number;
     clearDepth: boolean;
+
+    render(
+        renderer: WebGLRenderer,
+        _: WebGLMultipleRenderTargets | WebGLRenderTarget | null,
+        writeBuffer?: WebGLMultipleRenderTargets | WebGLRenderTarget,
```
Methuselah96, line 24: Shouldn't the parameters be `readBuffer` and `writeBuffer` (in that order)?

types/three/examples/jsm/postprocessing/Pass.d.ts
```diff
@@ -12,10 +12,10 @@ export class Pass {
     setSize(width: number, height: number): void;
     render(
         renderer: WebGLRenderer,
-        writeBuffer: WebGLRenderTarget,
-        readBuffer: WebGLRenderTarget,
-        deltaTime: number,
-        maskActive: boolean,
+        writeBuffer: WebGLMultipleRenderTargets | WebGLRenderTarget | null,
+        readBuffer?: WebGLMultipleRenderTargets | WebGLRenderTarget,
```
Methuselah96, line 16: I think the better fix would be to have `WebGLMultipleRenderTargets` extend `WebGLRenderTarget` like it does in the three.js source code.

Methuselah96 (changes requested):

types/three/examples/jsm/postprocessing/RenderPass.d.ts
repalash: One common way is to pass undefined in the constructor and assign the camera, scene before rendering. This way passes can be initialized before the scene/camera init/load.
But if its confusing I can remove this change.

types/three/examples/jsm/postprocessing/RenderPass.d.ts
repalash: In three.js [RenderPass.js](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/postprocessing/RenderPass.js#L57), the first parameter is ignored and the second parameter is used as writeBuffer. 
The autocomplete always creates confusion since the params are not `write, read` but `ignored, write`

types/three/examples/jsm/postprocessing/Pass.d.ts
repalash: I tried but since the type of texture is an object in `WebGLRenderTarget` and an array in `WebGLMultipleRenderTargets`, its not possible to inherit and change type of the property in typescript. This seemed to be the simplest solution for Passes atleast.
One way is to make `texture` to be `Texture|Texture[]` in `WebGLRenderTarget` but that might break a lot of projects and require special type casting at many places.
Another way is to make an interface like IWebGLRenderTarget and make both implement it. 

I am open to other ideas, would be good to fix type handling of `WebGLMultipleRenderTargets`.

Methuselah96: Makes sense, I've opened https://github.com/three-types/three-ts-types/pull/519 to resolve this.

types/three/examples/jsm/postprocessing/RenderPass.d.ts
Methuselah96: So the variables are misnamed in the three.js source code?

types/three/examples/jsm/postprocessing/RenderPass.d.ts
Methuselah96: I'm just concerned about the properties becoming optional since that would cause any access of those properties to assert that they're not `undefined` or check for it. I would feel better if there were some official three.js examples that show this pattern.

PR #468, opened 2023-05-25
Add event type as generic in EventDispatcher

repalash:
Add event type as generic in EventDispatcher

Add Support for event types in Object3D, Material, MeshBasicMaterial, MeshPhysicalMaterial, MeshStandardMaterial, ShaderMaterial, BufferGeometry, Camera.

### Why

Makes it easier to specify types of events like: `OrbitControls extends EventDispatcher<Event, 'change'|'end'|'start'>`.

Also adds generics for Core classes so that event types can be specified when inheriting the three.js classes.

### Checklist

-   [x] Ready to be merged

One of my first contributions here. Let me know if anything is wrong. Thanks.

### DIFF
--- a/types/three/src/cameras/Camera.d.ts
+++ b/types/three/src/cameras/Camera.d.ts
@@ -2,7 +2,7 @@ import { Matrix4 } from './../math/Matrix4';
 import { Vector3 } from './../math/Vector3';
 import { Object3D } from './../core/Object3D';
 import { Layers } from '../Three';
-import { BaseEvent, Event } from '../core/EventDispatcher'
+import { BaseEvent, Event } from '../core/EventDispatcher';
 
 /**
  * Abstract base class for cameras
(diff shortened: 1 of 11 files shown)


Methuselah96 (changes requested):
I think we want there to be a correlation between the event name and the type of the event, which is something @rafaelsc has started working on as part of https://github.com/three-types/three-ts-types/pull/369. I'm not sure if @rafaelsc is still working on that PR, but there are still some unresolved comments on that PR as well.

repalash:
Hello @Methuselah96, Ah, I didn't see that PR, will go through it.
In this way it's possible to infer the type from the event object type, but I added this as a shorthand for specifying the possible event types without the need for defining a separate typed event interface for each class. (Basically to get autocomplete in `addEventListener` and `dispatchEvent` functions for the types like this `OrbitControls extends EventDispatcher<Event, 'change'|'end'|'start'>`
 
We can hold on this till #369 is completed and see if it's useful then.

rafaelsc:
Sorry, @Methuselah96 I finally solved the conflict in my PR. I'm not using GitHub like I used before.

@repalash yes, I started working exactly your current PR, but in the end, I prefer the way that TS does for HTML events. Have autocomplete for the event name and the event properties.

Methuselah96:
Should be covered by https://github.com/three-types/three-ts-types/pull/369, can be reopened if not for some reason.

PR #469, opened 2023-05-26
BufferAttribute: add gpuType

CodyJasonBennett:
### Why

Follow-up to https://github.com/mrdoob/three.js/pull/26084.

### What

I've added `gpuType` to `BufferAttribute` to formalize its usage. `@link FloatType` doesn't work for the JSDoc though. Maybe we export type aliases for them -- `export type FloatType = typeof FloatType;`

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #479, opened 2023-06-18
add XRButton type

Yonet:
### Why

Add missing XRButton type

### What

Add new type

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/index.d.ts
+++ b/types/three/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for three 0.152
+// Type definitions for three 0.153
 // Project: https://threejs.org/
 // Definitions by: Josh Ellis <https://github.com/joshuaellis>
 //                 Nathan Bierema <https://github.com/Methuselah96>


types/three/examples/jsm/webxr/XRButton.d.ts
```diff
@@ -0,0 +1,5 @@
+import { WebGLRenderer } from '../../../src/Three';
+
+export namespace XRButton {
+    function createButton(renderer: WebGLRenderer): HTMLElement;
+}
```
Methuselah96: Can you make this a class with a static method?

Methuselah96 (changes requested):
Thanks for the PR!

Can you add a simple test for this?

types/three/examples/jsm/webxr/XRButton.d.ts
Yonet: Sure!

Yonet: @Methuselah96 I've changed it into a class and update the tests. When I run yarn test, I get the below error
'no-unnecessary-class  Every member of this class is static. Use namespaces or plain objects instead.'
Let me know how I should resolve.
Thanks!

Methuselah96: Huh, interesting. That seems like a strange lint rule to have for declaration files and seems to have more to do with original code. I would disable with a comment right above the class declaration with `// tslint:disable-next-line:no-unnecessary-class`.

PR #531, opened 2023-07-05
add(usdzexporter): USDZExporterOptions for quick look compatibility

ericbroberic:
### Why

These recent changes in USDZExporter (https://github.com/mrdoob/three.js/pull/26023) exposed a new optional boolean `quickLookCompatible` that defaults to `false`. Model Viewer, which uses TypeScript and QuickLook for AR on iOS needs to change that property to `true`. 

Model Viewer should be updated [here](https://github.com/google/model-viewer/pull/4314) or in a future PR.

### What

Added an optional interface as a second parameter to `USDZExporter`.

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #535, opened 2023-07-10
fix: ObjectLoader.d.ts

ybt-new:
Returns an Object instead of an Array
### Why
If you look at [ObjectLoader.parseTextures](https://github.com/mrdoob/three.js/blob/r154/src/loaders/ObjectLoader.js#L575-L682)

 you will notice that the returned textures is an object instead of an array

[ObjectLoader.parseMaterials](https://github.com/mrdoob/three.js/blob/r154/src/loaders/ObjectLoader.js#L311-L339) also returns an object


Methuselah96 (approved):
Thanks!

PR #536, opened 2023-07-12
delete some methods of WebGLState class

zack2012:
### Why
Delete some methods of WebGLState class since it does have these methods.

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #562, opened 2023-07-16
fix doc typo in Group.d.ts

ddnn55:
### Why

Surely the Group constructor creates a new Group rather than a new Bone.

### What

Replaced the word Bone with Group.

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #590, opened 2023-08-30
Change values parameter in KeyframeTrack class constructor to be optional

skyclouds2001:
### Why

Change values parameter in KeyframeTrack class constructor to be optional

### What

As metioned in the documentation of BooleanKeyframeTrack, parameter name and times are marked as a required parameter, while parameter values is marked as a optional parameter.

```markdown
## Constructor
#### BooleanKeyframeTrack( name : String, times : Array, values : Array )
name - (required) identifier for the KeyframeTrack.
times - (required) array of keyframe times.
values - values for the keyframes at the times specified.
```

Although in the documentation of KeyframeTrack, the requirement is not mentioned in the documentation, but it is the same with the documentation of BooleanKeyframeTrack.

And, in the source code of KeyframeTrack, passing undefined to parameter name and times will throw an error, while passing undefined to parameter values will not happen anything.

see https://threejs.org/docs/index.html#api/en/animation/KeyframeTrack
also see https://github.com/mrdoob/three.js/blob/master/src/animation/KeyframeTrack.js
see https://threejs.org/docs/index.html#api/en/animation/tracks/BooleanKeyframeTrack
also see https://github.com/mrdoob/three.js/blob/master/src/animation/tracks/BooleanKeyframeTrack.js
......

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/animation/KeyframeTrack.d.ts
+++ b/types/three/src/animation/KeyframeTrack.d.ts
@@ -8,10 +8,10 @@ export class KeyframeTrack {
     /**
      * @param name
      * @param times
-     * @param values
+     * @param [values]
      * @param [interpolation=THREE.InterpolateLinear]
      */
-    constructor(name: string, times: ArrayLike<number>, values: ArrayLike<any>, interpolation?: InterpolationModes);
+    constructor(name: string, times: ArrayLike<number>, values?: ArrayLike<any>, interpolation?: InterpolationModes);
 
     name: string;
     times: Float32Array;
--- a/types/three/src/animation/tracks/BooleanKeyframeTrack.d.ts
+++ b/types/three/src/animation/tracks/BooleanKeyframeTrack.d.ts
@@ -1,7 +1,7 @@
 import { KeyframeTrack } from './../KeyframeTrack.js';
 
 export class BooleanKeyframeTrack extends KeyframeTrack {
-    constructor(name: string, times: ArrayLike<number>, values: ArrayLike<any>);
+    constructor(name: string, times: ArrayLike<number>, values?: ArrayLike<any>);
 
     /**
      * @default 'bool'
--- a/types/three/src/animation/tracks/ColorKeyframeTrack.d.ts
+++ b/types/three/src/animation/tracks/ColorKeyframeTrack.d.ts
@@ -2,7 +2,7 @@ import { KeyframeTrack } from './../KeyframeTrack.js';
 import { InterpolationModes } from '../../constants.js';
 
 export class ColorKeyframeTrack extends KeyframeTrack {
-    constructor(name: string, times: ArrayLike<number>, values: ArrayLike<number>, interpolation?: InterpolationModes);
+    constructor(name: string, times: ArrayLike<number>, values?: ArrayLike<number>, interpolation?: InterpolationModes);
 
     /**
      * @default 'color'
--- a/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
+++ b/types/three/src/animation/tracks/NumberKeyframeTrack.d.ts
@@ -2,7 +2,7 @@ import { KeyframeTrack } from './../KeyframeTrack.js';
 import { InterpolationModes } from '../../constants.js';
 
 export class NumberKeyframeTrack extends KeyframeTrack {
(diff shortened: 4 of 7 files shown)


types/three/src/animation/KeyframeTrack.d.ts
```diff
@@ -8,10 +8,10 @@ export class KeyframeTrack {
     /**
      * @param name
      * @param times
-     * @param values
+     * @param [values]
      * @param [interpolation=THREE.InterpolateLinear]
      */
-    constructor(name: string, times: ArrayLike<number>, values: ArrayLike<any>, interpolation?: InterpolationModes);
+    constructor(name: string, times: ArrayLike<number>, values?: ArrayLike<any>, interpolation?: InterpolationModes);
```
Methuselah96, line 14: Does this mean that the `KeyframeTrack.values` should now be `Float32Array | undefined`?

Methuselah96 (changes requested):
Thanks for the PR, just a small note.

Methuselah96:
Do you mind explaining a little more when you wouldn't want to provide the `values` parameter to the `KeyframeTrack` class? I'm not too familiar with the usage of these classes. It seems like most of the methods in the class will throw an error if `values` was `undefined`, so I'm curious what your use case is.

types/three/src/animation/KeyframeTrack.d.ts
skyclouds2001: No, the `KeyframeTrack.values` does not require to be a Float32Array, any arraylike object is available.
In source code, it use [AnimationUtils.convertArray](https://github.com/mrdoob/three.js/blob/master/src/animation/KeyframeTrack.js#L21) to transform an array to be a Float32Array (by default) and any other types array according to `KeyframeTrack.prototype.ValueBufferType` (its default value is a Float32Array).

Methuselah96: Right I'm asking about [the property on the class](https://github.com/three-types/three-ts-types/blob/master/types/three/src/animation/KeyframeTrack.d.ts#L18) after it's been set by the constructor.

skyclouds2001:
> Do you mind explaining a little more when you wouldn't want to provide the `values` parameter to the `KeyframeTrack` class? I'm not too familiar with the usage of these classes. It seems like most of the methods in the class will throw an error if `values` was `undefined`, so I'm curious what your use case is.

I made an example in https://github.com/skyclouds2001/Frontend-Learning/blob/main/next-learning/demo.html.

And, it seems that passing undefined to parameter values will not throw an error when the class construct, but will throw an error if `KeyframeTrack.values` is undefined when in an animation.😂

```js
const opacityKeyframe = new THREE.NumberKeyframeTrack('.material.opacity', [0, 1, 2])

const clip = new THREE.AnimationClip('Action', 3, [visibleKeyframe, positionKeyframe, scaleKeyframe, colorKeyframe, opacityKeyframe, quaternionKeyframe])

const mixer = new THREE.AnimationMixer(mesh)

const action = mixer.clipAction(clip) // will throw an error here
action.play()
```

But it can set `KeyframeTrack.values` values with an Array or TypedArray after the class construct but before it is used to create aninmation.

```js
const opacityKeyframe = new THREE.NumberKeyframeTrack('.material.opacity', [0, 1, 2])
opacityKeyframe.values = [1, 0, 1] // or new Float32Array([1, 0, 1])

const clip = new THREE.AnimationClip('Action', 3, [visibleKeyframe, positionKeyframe, scaleKeyframe, colorKeyframe, opacityKeyframe, quaternionKeyframe])

const mixer = new THREE.AnimationMixer(mesh)

const action = mixer.clipAction(clip) // will not throw any error
action.play()
```

And as I tested, setting `KeyframeTrack.values` values must be executed before `action.play()` function is called.

Methuselah96:
Thanks for the explanation. I'm not sure that we should go with this change since it seems like you can't really use the class if `values` is undefined. It seems like more of an advanced use case to want to wait to set the `values` until after the class is constructed, and in order to work around the types in those use cases the user can either pass an empty array or pass `undefined` and cast it to an array.

skyclouds2001:
I think the better option right now is to keep it as it is.
And I'll open an issue in three.js to ask for help.

PR #614, opened 2023-09-25
Fix: change `Color.set()` override order to set ColorRepresentation function signature

michealparks:
### Why

Since r153 Typescript will throw errors when setting the Color property in a [Threlte](https://threlte.xyz) application, because Threlte will use the `Parameters` of `set()` to determine types.

### What

It was suggested in [this three-ts-types discussion](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/66812) to change override order to fix this issue.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/math/Color.d.ts
+++ b/types/three/src/math/Color.d.ts
@@ -204,8 +204,8 @@ export class Color {
      */
     b: number;
 
-    set(color: ColorRepresentation): this;
     set(r: number, g: number, b: number): this;
+    set(color: ColorRepresentation): this;
 
     /**
      * Sets this color's {@link r}, {@link g} and {@link b} components from the x, y, and z components of the specified


Methuselah96:
@michealparks Thanks for the PR!

Just to clarify, this would break the use-case of calling `<T.MeshStandardMaterial color={[255, 0, 0]} />`, correct? That may be the most pragmatic decision, given that I'm sure passing a string or `Color` is much more common, just want to make sure I'm understanding correctly that this does not completely solve the issue, it just solves the more common case.

michealparks:
Thank you for reviewing! You're correct. Although the issue won't be completely solved (as per your example) the following is much more common in Threlte:

```svelte
<T.MeshStandardMaterial color='turquoise' />

<!-- or -->
<!-- color is an instance of THREE.Color -->
<T.MeshStandardMaterial color={color} />
```

Methuselah96 (approved):
Thanks! This will get released with r157 at the end of the month.

PR #624, opened 2023-10-02
Fix(Matrix4.d.ts): Overloads is needed for `setPosition()`

ycw:
### Why

`setPosition()` accepts these [2 usages](https://github.com/mrdoob/three.js/blob/181e04eea8b569dc09048f9dc644310ed6b745a6/src/math/Matrix4.js#L469-L489):

```js
m4.setPosition(v) // Vector3
m4.setPosition(x,y,z) // 3 numbers, ALL REQUIRED
```

But current dts accepts extra usages which are illegal:

```js
m4.setPosition(x,y) // omit z
m4.setPosition(x) // omit y z
```

This PR forbids the illegal usages.

### Checklist

-   [x] Ready to be merged

### DIFF
--- a/types/three/src/math/Matrix4.d.ts
+++ b/types/three/src/math/Matrix4.d.ts
@@ -156,7 +156,8 @@ export class Matrix4 implements Matrix {
     /**
      * Sets the position component for this matrix from vector v.
      */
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

PR #627, opened 2023-10-11
Replace Color with ColorRepresentation

puxiao:
> Box3Helper.js

```
class Box3Helper extends LineSegments {
    constructor( box, color = 0xffff00 ) { ... }
}
```

But ...

```
new Box3Helper(box3, 0xffff00)

//Error: Argument of type 'number' is not assignable to parameter of type 'Color'.
```

----

<br>

> Color.js

```
set( r, g, b) {
    if ( g === undefined && b === undefined ) {
        const value = r;
        if ( value && value.isColor ) { 
        } else if ( typeof value === 'number' ) {
        } else if ( typeof value === 'string' ) {
        }
    }
    ...
 }
```

<br>

AxesHelper.js

```
setColors( xAxisColor, yAxisColor, zAxisColor ) {
    const color = new Color();
    color.set( xAxisColor );
    ...
}
```

<br>

So... 

> AxesHelper.d.ts

```diff
- setColors(xAxisColor: Color ...)
+ setColors(xAxisColor: ColorRepresentation ...)
```

### DIFF
--- a/types/three/src/helpers/AxesHelper.d.ts
+++ b/types/three/src/helpers/AxesHelper.d.ts
@@ -1,4 +1,4 @@
-import { Color } from '../math/Color.js';
+import { ColorRepresentation } from '../math/Color.js';
 import { LineSegments } from './../objects/LineSegments.js';
 
 /**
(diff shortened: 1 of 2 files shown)


Methuselah96 (approved):
Thanks!

PR #628, opened 2023-10-18
Add three/addons target for NPM

CodyJasonBennett:
### Why

In r158 dev I've added a `three/addons` target which exports all addons in https://github.com/mrdoob/three.js/pull/26910. We've tested tree-shaking and SSR upstream additionally.

### What

I've added `three/addons` to `package.json` as I did in three.js, and implemented those which were missing.

### Checklist

-   [x] Ready to be merged


types/three/examples/jsm/Addons.d.ts
```diff
@@ -0,0 +1,292 @@
...
+export * from './loaders/TTFLoader.js';
+export * from './loaders/TiltLoader.js';
+export * from './loaders/USDZLoader.js';
+export * from './loaders/VOXLoader.js';
+export * from './loaders/VRMLLoader.js';
+export * from './loaders/VTKLoader.js';
+export * from './loaders/XYZLoader.js';
+// export * from './loaders/lwo/IFFParser.js';
+// export * from './loaders/lwo/LWO2Parser.js';
+// export * from './loaders/lwo/LWO3Parser.js';
```
CodyJasonBennett: I don't believe these were meant to be exported. I'll check on this in three.js.

Methuselah96 (approved):
Thanks!

PR #629, opened 2023-10-20
Material: Allow `null` as value for `.clippingPlanes`

yfunk:
### Why
The default value of the property is `null`, so it should also be possible to assign `null` to it. 

Given that other properties of the `Material` class that default to `null` also allow this, it seems that this was simply an oversight. Previously, it was possible to assign `null` to it because the property was typed as `any` (until the typing was improved in [this commit](https://github.com/three-types/three-ts-types/commit/c7d06ae9cbeb8800008685dba371e4311fd39824)). 

### What

Adjust types to allow `null` as a possible value for the `clippingPlanes` property of `Material`

### Checklist

-   [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #660, opened 2023-11-02
BatchedMesh: geometry methods, error signatures

CodyJasonBennett:
Follow-up to https://github.com/three-types/three-ts-types/pull/648. Mirrors changes from https://github.com/mrdoob/three.js/pull/27078.


Methuselah96 (approved):
Thanks!

PR #679, opened 2023-11-15
[three] LineMaterial: fix incorrect type for color

sguimmara:
Changes the type of `color` in the constructor params from `number` to `ColorRepresentation`


Methuselah96 (approved):
Thanks!

PR #703, opened 2023-12-11
Fix KTX2Loader TS type to match the JS code

svignaud:
KTX2Loader TS type derives from CompresssedTextureLoader, and overrides its parse method. The actual JS code derives KTX2Loader from Loader and overrides its load method. This change will fix this mismatch.

### DIFF
--- a/types/three/examples/jsm/loaders/KTX2Loader.d.ts
+++ b/types/three/examples/jsm/loaders/KTX2Loader.d.ts
@@ -1,7 +1,7 @@
-import { LoadingManager, CompressedTextureLoader, CompressedTexture, WebGLRenderer } from '../../../src/Three.js';
+import { LoadingManager, Loader, CompressedTexture, WebGLRenderer } from '../../../src/Three.js';
 import WebGPURenderer from '../renderers/webgpu/WebGPURenderer.js';
 
-export class KTX2Loader extends CompressedTextureLoader {
+export class KTX2Loader extends Loader {
     constructor(manager?: LoadingManager);
 
     setTranscoderPath(path: string): KTX2Loader;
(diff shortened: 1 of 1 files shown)


types/three/examples/jsm/loaders/KTX2Loader.d.ts
```diff
@@ -1,17 +1,18 @@
-import { LoadingManager, CompressedTextureLoader, CompressedTexture, WebGLRenderer } from '../../../src/Three.js';
+import { LoadingManager, Loader, CompressedTexture, WebGLRenderer } from '../../../src/Three.js';
 import WebGPURenderer from '../renderers/webgpu/WebGPURenderer.js';
 
-export class KTX2Loader extends CompressedTextureLoader {
+export class KTX2Loader extends Loader {
```
Methuselah96: The `Loader` generic needs to be set for this to work:

```suggestion
export class KTX2Loader extends Loader<CompressedTexture> {
```

types/three/examples/jsm/loaders/KTX2Loader.d.ts
```diff
@@ -1,17 +1,18 @@
...
 
-    parse(
-        buffer: ArrayBuffer,
+    load(
+        url: string,
         onLoad: (texture: CompressedTexture) => void,
+        onProgress?: (event: ProgressEvent) => void,
         onError?: (event: ErrorEvent) => void,
-    ): KTX2Loader;
+    ): CompressedTexture;
```
Methuselah96: I think we can remove this method because it's inherited from `Loader`.

Methuselah96 (changes requested):
Thanks!

types/three/examples/jsm/loaders/KTX2Loader.d.ts
svignaud: Done

Methuselah96 (approved):
Thanks!

PR #705, opened 2023-12-12
Use `this` return type for `Matrix4` in-place functions

### LINKED ISSUE #704: Return `this` type instead of `Matrix4` for in-place mutations
# Objective

I would like to create a subclass of `Matrix4` to add my own custom methods while maintaining type safety.
However, when using functions like `.multiply`, the subclass is "lost", because it returns type `Matrix4`.
In reality, `.multiply` is an in-place operation which returns `this`, so the subclass is preserved.

## Example

```ts
export class Matrix4Ext extends Matrix4 {
  extension(): this {
    // -- Do stuff --
    return this;
  }
}

new Matrix4Ext()
  .multiply(new Matrix4())
  // Error: Property 'extension' does not exist on type 'Matrix4'.
  .extension();
```

# Solution

The type definitions should be adjusted to return `this` instead of `Matrix4` if the operation is in-place and `this` is returned.
Which type definitions have to be adjusted can be inferred from the [Three.js documentation on Matrix4](https://threejs.org/docs/#api/en/math/Matrix4) (using the `this` type).

TimJentzsch:
# Objective

Closes #704.

Consider the following example:

```ts
export class Matrix4Ext extends Matrix4 {
  extension(): this {
    // -- Do stuff --
    return this;
  }
}

new Matrix4Ext()
  .multiply(new Matrix4())
  // Error: Property 'extension' does not exist on type 'Matrix4'.
  .extension();
```

This example doesn't compile, because the subclass type is erased by the `Matrix4` return type of the `.multiply` function.

# Solution

Instead of `Matrix4`, we now use `this` as return type for functions which return `this`.
I used the [Matrix4 documentation on threejs.org](https://threejs.org/docs/#api/en/math/Matrix4) to determine which return types should be `this`.

### DIFF
--- a/types/three/src/math/Matrix4.d.ts
+++ b/types/three/src/math/Matrix4.d.ts
@@ -94,40 +94,40 @@ export class Matrix4 implements Matrix {
         n42: number,
         n43: number,
         n44: number,
-    ): Matrix4;
+    ): this;
 
     /**
      * Resets this matrix to identity.
      */
-    identity(): Matrix4;
+    identity(): this;
     clone(): Matrix4;
     copy(m: Matrix4): this;
-    copyPosition(m: Matrix4): Matrix4;
-    extractBasis(xAxis: Vector3, yAxis: Vector3, zAxis: Vector3): Matrix4;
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

PR #711, opened 2023-12-17
Add missing boolean readonly props to lights and materials

michealparks:
### Why
Currently it's difficult to type-narrow any class that does not have a typedef including an `is*` or `type` property that specifies a string constant as a value.

### What
This PR starts to address this problem by adding missing boolean readonly props to Lights and Materials.

### Checklist
- [x] Checked the target branch (current goes master, next goes dev)
- [ ] Added myself to contributors table
- [x] Ready to be merged


Methuselah96 (approved):
Thanks!

PR #719, opened 2023-12-20
Update Shader and UniformsUtils

vanruesc:
#### Shader

Added the missing `defines` property to the `Shader` interface which is often needed for use cases involving `Material.onBeforeCompile` hooks. The property is optional because not all materials have defines:

#### UniformsUtils

The params and return values of `UniformsUtils.clone()` and `UniformsUtils.merge()` now use concrete types to avoid troubles with `any` in user code.


types/three/src/renderers/shaders/UniformsUtils.d.ts
```diff
@@ -1,7 +1,8 @@
 import { UniformsGroup } from '../../core/UniformsGroup.js';
+import { IUniform } from './UniformsLib.js';
 
-export function cloneUniforms(uniforms_src: any): any;
-export function mergeUniforms(uniforms: any): any;
+export function cloneUniforms(uniformsSrc: { [uniform: string]: IUniform }): { [uniform: string]: IUniform };
+export function mergeUniforms(uniforms: { [uniform: string]: IUniform }): { [uniform: string]: IUniform };
```
Methuselah96: I think `mergeUniforms` should take an array of objects, right?

types/three/src/renderers/shaders/ShaderLib.d.ts
```diff
@@ -2,6 +2,7 @@ import { IUniform } from './UniformsLib.js';
 
 export interface Shader {
     uniforms: { [uniform: string]: IUniform };
+    defines?: { [define: string]: string | number | boolean };
```
Methuselah96: Hmm, I'm not sure this is the best place to fix this. I think instead the first parameter of `Material.onBeforeCompile` should be updated to be more accurate, and not just be a `Shader`.

It looks like that parameter is coming from [here](https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1645). The `parameters` object mostly [comes from](https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1603)  the [return value of `WebGLPrograms.getParameters`](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLPrograms.js#L177-L370) with [at least one modification](https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1641).

Methuselah96 (changes requested):

types/three/src/renderers/shaders/UniformsUtils.d.ts
vanruesc: Oh, you're right, good catch :+1:

types/three/src/renderers/shaders/ShaderLib.d.ts
vanruesc: True, should I replace `Shader` with a complete `WebGLProgramParameters` interface? Then we could also specify the return type of `WebGLPrograms.getParameters`.

Methuselah96:
Don't worry about the `test-examples` build failing too much, working with the examples testing is not well documented, so I can fix those.

types/three/src/renderers/shaders/ShaderLib.d.ts
Methuselah96: Yes, that would be great.

Methuselah96: Note that the argument passed into `Material.onBeforeCompile` has an additional `uniforms` property that is added after the `WebGLProgramParameters` is created, so I imagine we would also need a `WebGLProgramParametersWithUniforms` that extends from `WebGLProgramParameters` with the addition of a `uniforms` property.

types/three/src/renderers/shaders/UniformsUtils.d.ts
```diff
@@ -1,7 +1,8 @@
 import { UniformsGroup } from '../../core/UniformsGroup.js';
+import { IUniform } from './UniformsLib.js';
 
-export function cloneUniforms(uniforms_src: any): any;
-export function mergeUniforms(uniforms: any): any;
+export function cloneUniforms(uniformsSrc: { [uniform: string]: IUniform }): { [uniform: string]: IUniform };
```
Methuselah96: I think making this generic will make this better for users of this function:
```suggestion
export function cloneUniforms<T extends { [uniform: string]: IUniform }>(uniformsSrc: T): T;
```

Methuselah96 (changes requested):

types/three/src/renderers/shaders/ShaderLib.d.ts
```diff
@@ -1,9 +1,8 @@
+import { WebGLProgramParameters } from '../webgl/WebGLProgramParameters.js';
 import { IUniform } from './UniformsLib.js';
 
-export interface Shader {
+export interface Shader extends WebGLProgramParameters {
```
vanruesc: Should I rename `Shader` to `WebGLProgramParametersWithUniforms`? I think `Shader` works too as a parameter type.

types/three/src/renderers/shaders/ShaderLib.d.ts
vanruesc: Done. I've also specified the remaining types in `WebGLPrograms`. I moved the type of `WebGLLights.state` into a separate interface. Let me know if that's ok or if I should revert it.

types/three/src/renderers/webgl/WebGLPrograms.d.ts
```diff
@@ -20,9 +25,16 @@ export class WebGLPrograms {
...
-    getProgramCacheKey(parameters: any): string;
-    getUniforms(material: Material): object;
-    acquireProgram(parameters: any, cacheKey: string): WebGLProgram;
+    getParameters(
+        material: Material,
+        lights: WebGLLightsState,
+        shadows: Light[],
+        scene: Scene,
+        object: Object3D,
+    ): WebGLProgramParameters;
```
vanruesc, line 237: - `lights` comes from [here](https://github.com/mrdoob/three.js/blob/4acb78002425969f7c0cf46423ea5f728c41dfbc/src/renderers/WebGLRenderer.js#L1598) (the `state` property is passed into `getParameters`)
- `shadows` comes from [here](https://github.com/mrdoob/three.js/blob/4acb78002425969f7c0cf46423ea5f728c41dfbc/src/renderers/WebGLRenderer.js#L1599) and is populated with Lights [here](https://github.com/mrdoob/three.js/blob/4acb78002425969f7c0cf46423ea5f728c41dfbc/src/renderers/WebGLRenderer.js#L965) and [here](https://github.com/mrdoob/three.js/blob/4acb78002425969f7c0cf46423ea5f728c41dfbc/src/renderers/WebGLRenderer.js#L983).
- `object` comes from [here](https://github.com/mrdoob/three.js/blob/4acb78002425969f7c0cf46423ea5f728c41dfbc/src/renderers/WebGLRenderer.js#L999-L1025) and should be `Object3D`.

Methuselah96:
Nice work, thanks for taking that on! Will review soon.

types/three/src/renderers/shaders/ShaderLib.d.ts
Methuselah96: I was actually thinking we'd keep `Shader` as-is, since it's really just intended to describe the type of shaders in `ShaderLib`. Maybe it should be renamed to `ShaderLibShader` to make that more clear? I was thinking we'd export a `WebGLProgramParametersWithUniforms` interface from `WebGLRenderer`.

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
+import {
+    ColorSpace,
+    Combine,
+    DepthPackingStrategies,
+    GLSLVersion,
+    ShadowMapType,
+    ToneMapping,
+} from '../../constants.js';
+
+export interface WebGLProgramParameters {
```
Methuselah96: My preference is to avoid creating new declaration files that don't correspond to real JS files and to keep types as close to their source as possible. Can we move this interface to `WebGLPrograms`?

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+    isWebGL2: boolean;
+
+    shaderID: string;
+    shaderType: string;
+    shaderName: string;
+
+    vertexShader: string;
+    fragmentShader: string;
+    defines?: { [define: string]: string | number | boolean };
```
Methuselah96: nit: this will always exist, but it may be `undefined`
```suggestion
    defines: { [define: string]: string | number | boolean } | undefined;
```

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    customVertexShaderID?: string;
```
Methuselah96:
```suggestion
    customVertexShaderID: string | undefined;
```

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+    customFragmentShaderID?: string;
```
Methuselah96:
```suggestion
    customFragmentShaderID: string | undefined;
```

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    isRawShaderMaterial: boolean;
+    glslVersion?: GLSLVersion;
```
Methuselah96:
```suggestion
    glslVersion: GLSLVersion | null | undefined;
```

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+    instancing: boolean;
+    instancingColor: boolean;
+
+    supportsVertexTextures: boolean;
+    outputColorSpace: ColorSpace;
+
+    map: boolean;
+    matcap: boolean;
+    envMap: boolean;
+    envMapMode: boolean;
```
Methuselah96: Should this be `Mapping | false`?

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    gradientMap: boolean;
+
+    opaque: boolean;
+
+    alphaMap: boolean;
+    alphaTest: boolean;
+    alphaHash: boolean;
+
+    combine: Combine;
```
Methuselah96: `Combine | undefined`?

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+    sheenRoughnessMapUv: boolean;
+
+    specularMapUv: boolean;
+    specularColorMapUv: boolean;
+    specularIntensityMapUv: boolean;
+
+    transmissionMapUv: boolean;
+    thicknessMapUv: boolean;
+
+    alphaMapUv: boolean;
```
Methuselah96: `string | false`?

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    decodeVideoTexture: boolean;
+
+    premultipliedAlpha: boolean;
+
+    doubleSided: boolean;
+    flipSided: boolean;
+
+    useDepthPacking: boolean;
+    depthPacking: DepthPackingStrategies;
```
Methuselah96: `DepthPackingStrategies | 0`

types/three/src/renderers/webgl/WebGLProgramParameters.d.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    index0AttributeName: string;
```
Methuselah96: `string | undefined`

Methuselah96 (changes requested):

types/three/src/renderers/shaders/ShaderLib.d.ts
vanruesc: I've exported `WebGLProgramParametersWithUniforms` from `WebGLPrograms` since it's used there as a param type for `acquireProgram`. Is there a specific reason why it should be in `WebGLRenderer`?

Methuselah96: No, that should be fine, thanks.

Methuselah96 (approved):
Thanks, that was more work than anticipated!

vanruesc:
Yeah, and those mixed types were throwing me off :sweat_smile: Thanks for the help!

PR #728, opened 2023-12-23
GLTFLoaderPlugin: add constructor and required name getter

hybridherbst:
The constructor [always gets the parser passed](https://github.com/mrdoob/three.js/blob/299f53ff4e65041a95b13c98b8da24a47cb877ea/examples/jsm/loaders/GLTFLoader.js#L376), and [the name getter is required](https://github.com/mrdoob/three.js/blob/299f53ff4e65041a95b13c98b8da24a47cb877ea/examples/jsm/loaders/GLTFLoader.js#L378).


Methuselah96:
The parser is already getting passed via the callback in the `register` and `unregister`, so I don't think the constructor is necessary.

```ts
export class GLTFLoader extends Loader<GLTF> {
    // ...
    register(callback: (parser: GLTFParser) => GLTFLoaderPlugin): GLTFLoader;
    unregister(callback: (parser: GLTFParser) => GLTFLoaderPlugin): GLTFLoader;
    // ...
```

Methuselah96 (changes requested):
Thanks for the PR!

The parser is already getting passed via the callback in the `register` and `unregister`, so I don't think adding a constructor is correct.

```ts
export class GLTFLoader extends Loader<GLTF> {
    // ...
    register(callback: (parser: GLTFParser) => GLTFLoaderPlugin): GLTFLoader;
    unregister(callback: (parser: GLTFParser) => GLTFLoaderPlugin): GLTFLoader;
    // ...
}
```

types/three/examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -131,6 +131,8 @@ export class GLTFParser {
 }
 
 export interface GLTFLoaderPlugin {
+    constructor(parser: GLTFParser);
+    get name(): string;
```
Methuselah96: nit: my personal preference is to use a normal property when it could be either, unless I'm missing a reason to do it as a getter.
```suggestion
    readonly name: string;
```

Methuselah96 (changes requested):

hybridherbst (commented):
remove constructor again, change name to readonly prop

hybridherbst:
Thanks! I think you're right about the constructor, and I changed to the readonly prop.

PR #733, opened 2023-12-25
fix (Nodes): fix type of `clamp` and `saturate`

0b5vr:
### What

- `clamp` can be unary or binary
  - See: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/math/MathNode.js#L301
  - I have no confidence of the type name `UnaryOrBinaryOrTernary`. Would you come up with a better name?
- `saturate` is unary

### DIFF
--- a/types/three/examples/jsm/nodes/math/MathNode.d.ts
+++ b/types/three/examples/jsm/nodes/math/MathNode.d.ts
@@ -183,7 +183,11 @@ export type Ternary = (
 ) => ShaderNodeObject<MathNode>;
 
 export const mix: Ternary;
-export const clamp: (a: NodeRepresentation, b: NodeRepresentation, c: NodeRepresentation) => ShaderNodeObject<MathNode>;
+export const clamp: (
+    a: NodeRepresentation,
+    b?: NodeRepresentation,
+    c?: NodeRepresentation,
+) => ShaderNodeObject<MathNode>;
 export const saturate: Unary;
 export const refract: Ternary;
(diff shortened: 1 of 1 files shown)


Methuselah96:
Looks like you may have been working on this change around the same as I made some changes to `Ternary`. Does [this change](https://github.com/three-types/three-ts-types/pull/731/files#diff-a05f8ef552222803775c3f53a2d3808e85e39c93f37608f5e087202dd7b1a2a3R181-R182) not fix this? Or do we know for sure that `clamp` and `saturate` are special in that they allow missing arguments?

0b5vr:
> Looks like you may have been working on this change around the same as I made some changes to Ternary. Does [this change](https://github.com/three-types/three-ts-types/pull/731/files#diff-a05f8ef552222803775c3f53a2d3808e85e39c93f37608f5e087202dd7b1a2a3R181-R182) not fix this? Or do we know for sure that clamp and saturate are special in that they allow missing arguments?

I'm sure other `Ternary` nodes like `refract`, `smoothstep`, or `mix` doesn't work properly without the second or the third arguments. Tested on https://threejs.org/examples/?q=tsl#webgpu_tsl_editor

`saturate` is definitely `Unary`, it doesn't have the second or the third arguments.
`clamp` has default parameter for the second and the third arguments as I indicated in the link.

Methuselah96:
Good call, I've updated those to be required again. Not sure why my mind thought they should all be that way.

Methuselah96 (approved):

PR #735, opened 2023-12-26
feat (Nodes): type `tslFn` and several functions that use tslFn

0b5vr:
### What

Type `tslFn` and several functions that use tslFn.

Not exhaustive.
Have very low confidence whether the typing direction is correct.

- `tslFn`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/shadernode/ShaderNode.js#L485-L520
- `BlendModeNode.js`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/display/BlendModeNode.js
- `BRDF_GGX`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/BRDF_GGX.js
- `BRDF_Lambert`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/BRDF_Lambert.js
- `BRDF_Sheen`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/BRDF_Sheen.js
- `D_GGX`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/D_GGX.js
- `DFGApprox`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/DFGApprox.js
- `F_Schlick`: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/functions/BSDF/F_Schlick.js

### DIFF
--- a/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
+++ b/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
@@ -164,10 +164,8 @@ export function nodeImmutable<T>(
 ): ShaderNodeObject<ConstructedNode<T>>;
 
 export function tslFn<T, R extends Node = ShaderNodeObject<Node>>(
-    jsFunc: (args: T) => R
-): T extends AnyObject
-    ? (args: T) => R
-    : () => R;
+    jsFunc: (args: T) => R,
+): T extends AnyObject ? (args: T) => R : () => R;
 
 export function append(node: Node): Node;
 
--- a/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
+++ b/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
@@ -1,5 +1,5 @@
 import Node from '../core/Node.js';
-import { NodeTypeOption, SwizzleOption } from '../core/constants.js';
+import { AnyObject, NodeTypeOption, SwizzleOption } from '../core/constants.js';
 import ConstNode from '../core/ConstNode.js';
 import NodeBuilder from '../core/NodeBuilder.js';
 
@@ -162,6 +162,12 @@ export function nodeImmutable<T>(
     nodeClass: T,
     ...params: ProxiedTuple<GetConstructors<T>>
 ): ShaderNodeObject<ConstructedNode<T>>;
+
+export function tslFn<T, R extends Node = ShaderNodeObject<Node>>(
+    jsFunc: (args: T) => R
+): T extends AnyObject
+    ? (args: T) => R
+    : () => R;
 
 export function append(node: Node): Node;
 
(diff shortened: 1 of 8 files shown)


types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
```diff
@@ -163,6 +163,10 @@ export function nodeImmutable<T>(
     ...params: ProxiedTuple<GetConstructors<T>>
 ): ShaderNodeObject<ConstructedNode<T>>;
 
+export function tslFn<T, R extends Node = ShaderNodeObject<Node>>(
+    jsFunc: (args: T) => R,
+): T extends AnyObject ? (args: T) => R : () => R;
```
Methuselah96: Can you explain the return type? I see that `tslFn` does fancy things with the first argument, but not seeing a clear explanation for why we would just remove the first argument if it's not an object. I would be inclined to just return `(args: T) => R` unless I'm missing something.

Methuselah96 (changes requested):

types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts
0b5vr: `tslFn` can define functions without arguments, and it's used in the way in several nodes. `GaussianBlurNode` for example.

Definition: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/display/GaussianBlurNode.js#L103
Use: https://github.com/mrdoob/three.js/blob/r160/examples/jsm/nodes/display/GaussianBlurNode.js#L136

In this case, the user wants to call the function without argument, and the ternary return type achieves such cases.

To clarify the use, changing `T` to `T extends AnyObject | unknown` might help.

Methuselah96: Thanks, that's helpful. Can you add simple tests to `test/unit/examples/jsm/nodes/shaderNode/ShaderNode.ts` for both the single argument case and the no argument case? I feel like there might be a better way to type this, but I want to see this way functioning first before suggesting alternatives.

Methuselah96: One idea that crossed my mind is using function overloads instead. Something like:

```ts
export function tslFn<T, R extends Node = ShaderNodeObject<Node>>(
    jsFunc: (args: T) => R,
): (args: T) => R;
export function tslFn<R extends Node = ShaderNodeObject<Node>>(
    jsFunc: () => R,
): () => R;
```

Another idea might be to make the function itself the generic, but that might make it harder to manipulate the argument if needed.

0b5vr: overloading looks great. I'll go with it. Thank you for the suggestion!

0b5vr: added a test as well

Methuselah96 (approved):
Thanks!

PR #736, opened 2023-12-26
WebGLPrograms: Update `fogExp2` parameter

vanruesc:
This change is for `three` r161, see https://github.com/mrdoob/three.js/pull/27430.


0b5vr:
@vanruesc good one! the PR should be directed to `dev` instead of `master` though, since it's a change for the next release of Three.js.

vanruesc:
Is this something I can change? I don't see a `dev` branch :thinking:

Methuselah96:
@0b5vr I actually got rid of the distinction in https://github.com/three-types/three-ts-types/pull/662, sorry for the confusion. Feel free to advocate to bring it back, but most of the time we just ended up publishing once a month anyway.

0b5vr:
> @0b5vr I actually got rid of the distinction in https://github.com/three-types/three-ts-types/pull/662, sorry for the confusion. Feel free to advocate to bring it back, but most of the time we just ended up publishing once a month anyway.

got it! I think that's a reasonable change.

@vanruesc seems we are going to go with `master` even when we are targeting the next update. sorry for the confusion 🙇

0b5vr (approved):

PR #737, opened 2023-12-28
Allow null as DataArrayTexture data

Oletus:
This should be allowed and other DataTexture types already allow it.


Methuselah96 (approved):
Thanks!

PR #739, opened 2024-01-02
Event maps on objects

sguimmara:
Add generic `TEventMap` parameters (with appropriate default value) to all objects under `src/objects` to be able to subclass them and specify event maps.


sguimmara:
The build error seems unrelated to this PR, as I can reproduce it on `master`

Methuselah96 (approved):
Thanks!

PR #741, opened 2024-01-09
Add generic Renderer type and Color4

schwyzl:
Added type definitions for the new generic Renderer type for the WebGPU Renderer. Also added the new Color4 type referenced by Renderer.

### DIFF
--- a/types/three/examples/jsm/renderers/common/Color4.d.ts
+++ b/types/three/examples/jsm/renderers/common/Color4.d.ts
@@ -5,9 +5,7 @@ export type Color4Representation = ColorRepresentation | Color4;
 export default class Color4 extends Color {
     constructor(r: number, g: number, b: number, a?: number);
 
-    set(...args: [color: Color4Representation] | [r: number, g: number, b: number, a?: number]): this;
-
-    copy(color: this): this;
+    set(...args: [color: ColorRepresentation] | [r: number, g: number, b: number, a?: number]): this;
 
     clone(): this;
 }
--- a/types/three/examples/jsm/renderers/common/Color4.d.ts
+++ b/types/three/examples/jsm/renderers/common/Color4.d.ts
@@ -1,0 +1,14 @@
+import { Color } from 'three';
+import { ColorRepresentation } from '../../../../src/math/Color';
+
+export type Color4Representation = ColorRepresentation | Color4;
+
+export default class Color4 extends Color {
+    constructor(r: number, g: number, b: number, a?: number);
+
+    set(...args: [color: Color4Representation] | [r: number, g: number, b: number, a?: number]): this;
+
+    copy(color: this): this;
+
+    clone(): this;
+}
--- a/types/three/examples/jsm/renderers/common/Color4.d.ts
+++ b/types/three/examples/jsm/renderers/common/Color4.d.ts
@@ -1,5 +1,4 @@
-import { Color } from 'three';
-import { ColorRepresentation } from '../../../../src/math/Color';
+import { Color, ColorRepresentation } from '../../../../src/math/Color';
 
 export type Color4Representation = ColorRepresentation | Color4;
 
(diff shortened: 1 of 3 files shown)


types/three/examples/jsm/renderers/common/Color4.d.ts
```diff
@@ -0,0 +1,13 @@
+import { Color, ColorRepresentation } from '../../../../src/math/Color.ts';
+
+export type Color4Representation = ColorRepresentation | Color4;
+
+export default class Color4 extends Color {
+    constructor(r: number, g: number, b: number, a?: number);
+
+    set(...args: [color: Color4Representation] | [r: number, g: number, b: number, a?: number]): this;
```
Methuselah96: It doesn't seem like `Color4.set` can handle a `Color4` correctly. Maybe we should just stick with `ColorRepresentation` instead of `Color4Representation` here?

types/three/src/math/Color.d.ts
```diff
@@ -255,7 +255,7 @@ export class Color {
      * Copies given color.
      * @param color Color to copy.
      */
-    copy(color: Color): this;
+    copy(color: this): this;
```
Methuselah96, line 258: Is there a benefit to changing the parameter to `this`? Seems like it could still just take a `Color` (and `Color4.copy()` could take a `Color4`)?

types/three/examples/jsm/renderers/common/Renderer.d.ts
```diff
@@ -1,27 +1,83 @@
...
 
+/**
+ * Generic Renderer interface containing either a WebGL or WebGPU backend.
+ */
 export default class Renderer {
-    isRenderer: true;
+    /**
+     * @default true
+     */
+    isRenderer: boolean;
```
Methuselah96: We usually type these as readonly since they're not supposed to be modified:

```
readonly isRenderer: true;
```

types/three/examples/jsm/renderers/common/Renderer.d.ts
```diff
@@ -36,7 +92,121 @@ export default class Renderer {
...
+
+    /**
+     * Sets the custom opaque sort function for the RenderLists. Pass null to use the default painterSortStable function.
+     */
+    setOpaqueSort(method: (a: any, b: any) => number): void;
+
+    /**
+     * Sets the custom transparent sort function for the RenderLists. Pass null to use the default reversePainterSortStable function.
+     */
+    setTransparentSort(method: (a: any, b: any) => number): void;
```
Methuselah96: Can we use `unknown` instead of `any`?

Methuselah96 (changes requested):
Thanks for working on this!

types/three/src/math/Color.d.ts
schwyzl: The benefit was that it wouldn't need redefinition in Color4, but I neglected to remove the definition from Color4.

schwyzl: Would you prefer me to make the change and remove the definition from color4 or define in both color and color4?

types/three/examples/jsm/renderers/common/Renderer.d.ts
schwyzl: yep, I can amend that no problem

types/three/examples/jsm/renderers/common/Color4.d.ts
schwyzl: happy to try

types/three/src/math/Color.d.ts
Methuselah96: I'm fine either way, I can't think of a situation where it would make a difference.

types/three/examples/jsm/renderers/common/Color4.d.ts
```diff
@@ -0,0 +1,11 @@
+import { Color, ColorRepresentation } from '../../../../src/math/Color.ts';
+
+export type Color4Representation = ColorRepresentation | Color4;
```
Methuselah96: Looks like this type is unused now.

Methuselah96 (changes requested):

types/three/examples/jsm/renderers/common/Color4.d.ts
schwyzl: True, I will remove

Methuselah96 (approved):
Thanks!

PR #748, opened 2024-01-16
Fixed wrong type in SSAOPass constructor

j-zarzoso:
Type for member 'height' is different than type defined in the constructor.


Methuselah96 (approved):
Thanks!

PR #770, opened 2024-01-22
add missing compute declaration to renderer (and others)

schwyzl:


types/three/examples/jsm/renderers/common/Renderer.d.ts
```diff
@@ -214,4 +229,25 @@ export default class Renderer {
...
+    /**
+     * Runs a compute pipeline
+     */
+    compute(computeNodes: ComputeNode | ComputeNode[]): Promise<void>;
+
+    hasFeature(name: string): boolean;
+
+    copyFramebufferToTexture(framebufferTexture: FramebufferTexture): void;
+
+    readRenderTargetPixelsAsync(renderTarget: RenderTarget, x: number, y: number, width: number, height: number): void;
```
Methuselah96: Seems like this should return a promise that resolves to some sort of typed array?

Methuselah96 (changes requested):

types/three/examples/jsm/renderers/common/Renderer.d.ts
schwyzl: Oh good catch. will take a look

Methuselah96 (approved):
Thanks!

PR #777, opened 2024-01-23
PerspectiveCamera: getViewBounds(), getViewSize()

Methuselah96:
- https://github.com/mrdoob/three.js/pull/27574
- https://github.com/mrdoob/three.js/pull/27605

### DIFF
--- a/types/three/src/cameras/PerspectiveCamera.d.ts
+++ b/types/three/src/cameras/PerspectiveCamera.d.ts
@@ -1,6 +1,5 @@
 import { Camera } from './Camera.js';
-import { Vector3 } from '../math/Vector3';
-import { Vector2 } from '../math/Vector2';
+import { Vector2 } from '../math/Vector2.js';
 
 /**
  * Camera that uses {@link https://en.wikipedia.org/wiki/Perspective_(graphical) | perspective projection}.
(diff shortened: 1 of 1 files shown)


types/three/src/cameras/PerspectiveCamera.d.ts
```diff
@@ -148,6 +149,18 @@ export class PerspectiveCamera extends Camera {
...
+     * Computes the 2D bounds of the camera's viewable rectangle at a given distance along the viewing direction.
+     * Sets minTarget and maxTarget to the coordinates of the lower-left and upper-right corners of the view rectangle.
+     */
+    getViewBounds(distance: number, minTarget: Vector2, maxTarget: Vector2): void;
+
+    /**
+     * Computes the width and height of the camera's viewable rectangle at a given distance along the viewing direction.
+     * Copies the result into the target Vector2, where x is width and y is height.
+     */
+    getViewSize(distance: number, minTarget: Vector2, maxTarget: Vector2): void;
```
linbingquan, line 162: @Methuselah96 Here it should be

```ts
getViewSize(distance: number, target: Vector2): Vector2;
```

Methuselah96: Good catch, thanks. Fixed in https://github.com/three-types/three-ts-types/pull/800.

PR #798, opened 2024-01-31
Use dprint for formatting examples testing

Methuselah96:
Follow-up to https://github.com/three-types/three-ts-types/pull/797.


Methuselah96:
Patch needs to be fixed to not have formatting noise.

PR #807, opened 2024-02-03
Timer: add timestamp param and FixedTimer

vanruesc:
This PR adds the optional `timestamp` parameter to the `update` method (https://github.com/mrdoob/three.js/pull/27421). The fixed delta logic has also been moved into a separate `FixedTimer` class (https://github.com/mrdoob/three.js/pull/27423).


Methuselah96 (approved):
Thanks! Not sure how I missed these changes.

PR #808, opened 2024-02-04
Add onBeforeCompile to MaterialParameters

vaneenige:
This PR includes the `onBeforeCompile` callback type to `MaterialParameters`.

Currently the type is valid if you create a material and call `onBeforeCompile` on it afterwards, but it can also be provided during creation (for which now the property type is unknown).

```
Object literal may only specify known properties, and 'onBeforeCompile' does not exist in type 'MeshDepthMaterialParameters'.
```

### DIFF
--- a/types/three/src/materials/Material.d.ts
+++ b/types/three/src/materials/Material.d.ts
@@ -37,6 +37,7 @@ export interface MaterialParameters {
     depthTest?: boolean | undefined;
     depthWrite?: boolean | undefined;
     name?: string | undefined;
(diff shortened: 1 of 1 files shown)


Methuselah96:
Hey @vaneenige!

Unfortunately, I'm not sure if the types should support this.

The docs for each material mention that the `parameters` parameter is "an object with one or more properties defining the material's appearance." The `onBeforeCompile` is not a property, its a method. I have also not seen any example code showing `onBeforeCompile` being passed as a parameter.

Since it's not clear to me whether this is intended/supported behavior, I'd rather not add it to the types. Presumably we would also have to add all other methods to the parameter types as well if we went down this route.

If you still want to do this you could create your own parameters type or cast your parameters.

PR #840, opened 2024-02-26
[@type/three] Fixed import extension.

jpremkumar:
Exception on import: An import path cannot end with a '.ts' extension. Consider importing `../../src/math/color.js instead.


Methuselah96 (approved):
Thanks!

PR #850, opened 2024-02-29
Use "module": "node16"

Methuselah96:
Apparently `three` was listed as not working with `node16` in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68529, which seems odd since we use `node16` in our examples testing. This PR is to figure out what's going on.


Methuselah96:
Looks like we need to wait for TypeScript 4.6 support to get dropped to fix some of these errors.

PR #890, opened 2024-04-08
feat: Support the latest MaterialNode exports as of r163

0b5vr:
This PR supports the latest MaterialNode scopes and exports as of r163.

Ref: https://github.com/mrdoob/three.js/blob/r163/examples/jsm/nodes/accessors/MaterialNode.js


Methuselah96 (approved):
Thanks!

PR #892, opened 2024-04-13
SelectionBox: add instance property

edsunman:
It looks like the SelectionBox class has an `instances` property that is missing in SelectionBox.d.ts

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/interactive/SelectionBox.js

I'm relatively new to typescript so hopefully that type definition is correct!

### DIFF
--- a/types/three/examples/jsm/interactive/SelectionBox.d.ts
+++ b/types/three/examples/jsm/interactive/SelectionBox.d.ts
@@ -8,7 +8,7 @@ export class SelectionBox {
     endPoint: Vector3;
     scene: Scene;
     startPoint: Vector3;
-    instances: {} | {[key:string]:number[]};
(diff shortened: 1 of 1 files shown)


types/three/examples/jsm/interactive/SelectionBox.d.ts
```diff
@@ -8,6 +8,7 @@ export class SelectionBox {
     endPoint: Vector3;
     scene: Scene;
     startPoint: Vector3;
+    instances: {} | {[key:string]:number[]};
```
Methuselah96:
```suggestion
    instances: Record<string, number[]>;
```

Methuselah96 (approved):
Thanks for the PR! I tweaked the type a little bit, but otherwise looks good.

PR #937, opened 2024-05-14
Implement *Assign

### LINKED ISSUE #936: NodeMaterial: `addAssign` and other "operate and assign" methods are missing
### Description

We currently don't have type definitions for `ShaderNodeObject.addAssign` and other `***Assign` methods.
We need to add them.

### Workaround

We can use `.assign` and the desired operation separately instead. Though it's slightly redundant to write, the code inside `***Assign` also does the same so there is almost no performance debt.

```ts
node.addAssign(other); // unable to use in TS

node.assign(node.add(other)); // has the same effect
```

Alternatively, you can patch the [ShaderNode.d.ts](https://github.com/three-types/three-ts-types/blob/r164/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts) by using solutions like [patch-package](https://github.com/ds300/patch-package) if you still want to use the `addAssign`:

```ts
export interface NodeElements {
    // ...

    addAssign: (
        nodeA: NodeRepresentation,
        nodeB: NodeRepresentation,
        ...others: NodeRepresentation[]
    ) => ShaderNodeObject<Node>;
}
```

### Source

Here is the source on the Three.js side:

https://github.com/mrdoob/three.js/blob/r164/examples/jsm/nodes/shadernode/ShaderNode.js#L67

### Possible solution

I'm 75% sure that we are going to add something in [ShaderNode.d.ts](https://github.com/three-types/three-ts-types/blob/r164/types/three/examples/jsm/nodes/shadernode/ShaderNode.d.ts).

Methuselah96:
Fixes #936


0b5vr:
Thank you! I will try if it's working properly.

PR #938, opened 2024-05-15
USDZExporter: Add ar and includeAnchoringProperties options

vincentfretin:
The `includeAnchoringProperties` option is to be included in r165, see https://github.com/mrdoob/three.js/pull/28363


Methuselah96 (approved):
Thanks!

PR #941, opened 2024-05-17
Add missing mode argument for TransformControls.d.ts events

s-rigaud:
The `TransformControls` actually returns the mode each time `mouseDown` and `mouseUp` events are triggered (same behaviour for older library versions).

See the full implementation here : https://github.com/mrdoob/three.js/blob/master/examples/jsm/controls/TransformControls.js


Methuselah96 (approved):
Thanks!

PR #943, opened 2024-05-17
Raycaster: add ability to stop raycast traversal

Methuselah96:
https://github.com/mrdoob/three.js/pull/27709


Methuselah96:
I guess you can override it to return `boolean` if you want. Not sure there's actually any type change to make here.

PR #1013, opened 2024-06-14
LineMaterialParameters: extends ShaderMaterialParameters not MaterialParameters

puxiao:
```
class LineMaterial extends ShaderMaterial { ... }
```

So,  the constructor parameters should also extends `ShaderMaterialParameters` , not `MaterialParameters` .

<br>

```
export interface ShaderMaterialParameters extends MaterialParameters {
    ...
    linewidth?: number | undefined;
    wireframe?: boolean | undefined;
    ...
}
```

So, We can remove linewidth and wireframe from LineMaterialParameters.


Methuselah96 (approved):
Thanks!

PR #1017, opened 2024-06-17
WebGPURenderer: Add types for WebGPUAttributeUtils and WebGPUBindingUtils.

cmhhelgeson:
Add types for WebGPUAttributeUtils and WebGPUBindingUtils, both of which can mostly be typed without explicit access to WebGPU types from the webgpu/types library.


Methuselah96 (commented):
Seems like return types are missing on the methods?

types/three/examples/jsm/renderers/webgpu/utils/WebGPUAttributeUtils.d.ts
```diff
@@ -0,0 +1,14 @@
+import { BufferAttribute } from "three/src/Three.js";
```
Methuselah96:
```suggestion
import { BufferAttribute } from "three";
```

PR #1024, opened 2024-06-18
Fix type of `CompressedTexture.mipmaps`

HunterLarco:
Currently `CompressedTexture` and `CompressedArrayTexture` represent mipmaps using the browser's [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData). This is actually inaccurate to three.js's implementation in a few ways:

1. [`ImageData#colorSpace`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/colorSpace) is unused by three.js.
2. [`ImageData#data`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data) is the incorrect type. It is `Uint8ClampedArray` however, using `Uint8ClampedArray` will _sometimes_ cause runtime breakages in three.js (see below for details)

Of these two, (2) is the real problem. [`ImageData#data`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data) is defined as `Uint8ClampedArray`, however, mipmap data is [regularly used by threejs](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLTextures.js#L795) in [`texSubImage2D`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D) and [`compressedTexSubImage2D`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/compressedTexSubImage2D) which both require that the source data array view matches the texture type. For example, if I have a texture using `RGIntegerFormat` with `FloatType` then `compressedTexSubImage2D` must be passed a `Float32Array`. As currently written, clients would be forced to use `Uint8ClampedArray` which will cause a runtime error due to misalignment between the source data and texture type. This is documented by [`texSubImage2D`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D) which specifies

> One of the following objects can be used as a pixel source for the texture:
> 
> - [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) (Must be used if type is gl.UNSIGNED_BYTE)
> - [Uint16Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint16Array) (Must be used if type is either gl.UNSIGNED_SHORT_5_6_5, gl.UNSIGNED_SHORT_4_4_4_4, gl.UNSIGNED_SHORT_5_5_5_1, or ext.HALF_FLOAT_OES)
> - [Float32Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array) (Must be used if type is gl.FLOAT)

The fix here is to replace the current usage of `ImageData` with a new type `MipmapImageData` which better aligns to three.js's internal usage, and importantly allows the `data` field to be populated by any `TypedArray`.

### DIFF
--- a/types/three/src/textures/CompressedArrayTexture.d.ts
+++ b/types/three/src/textures/CompressedArrayTexture.d.ts
@@ -1,6 +1,5 @@
 import { CompressedPixelFormat, TextureDataType, Wrapping } from "../constants.js";
-import { CompressedTexture } from "./CompressedTexture.js";
-import { MipmapImageData } from "./types.js";
+import { CompressedTexture, CompressedTextureMipmap } from "./CompressedTexture.js";
 
 /**
  * Creates an texture 2D array based on data in compressed form, for example from a
@@ -38,8 +37,8 @@ export class CompressedArrayTexture extends CompressedTexture {
 
     /**
      * Create a new instance of {@link CompressedArrayTexture}
-     * @param mipmaps The mipmaps array should contain objects with data, width and height.
-     * The mipmaps should be of the correct {@link format} and {@link type}. See {@link THREE.mipmaps}.
+     * @param mipmaps The mipmaps array should contain objects with data, width and height. The mipmaps should be of the
+     * correct format and type.
      * @param width The width of the biggest mipmap.
      * @param height The height of the biggest mipmap.
      * @param depth The number of layers of the 2D array texture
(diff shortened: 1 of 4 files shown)


Methuselah96 (approved):
Thanks! I did some cleanup to the PR, hopefully while preserving the intent of your changes.

HunterLarco:
This looks great! Thanks for tidying up the PR

PR #1025, opened 2024-06-18
Strongly type Texture.mipmaps

Methuselah96:
Typing `Texture.mipmaps` as `any[]` is too loose.


types/three/src/textures/CompressedTexture.d.ts
```diff
@@ -72,7 +72,7 @@ export class CompressedTexture extends Texture {
      *  The mipmaps array should contain objects with data, width and height. The mipmaps should be of the correct
      *  format and type.
      */
-    mipmaps: CompressedTextureMipmap[];
+    mipmaps: CompressedTextureMipmap[] | undefined;
```
HunterLarco, line 75: I'm not sure mipmaps actually can be undefined here?

https://github.com/mrdoob/three.js/blob/dev/src/textures/CompressedTexture.js#L12

Methuselah96: `CompressedCubeTexture` sets it to `undefined` [here](https://github.com/mrdoob/three.js/blob/dev/src/textures/CompressedCubeTexture.js#L8). 😢

HunterLarco: That is deeply unfortunate 😢 but nice catch!

PR #1071, opened 2024-07-17
Three json

Hoodgail:
Minor fixes

PR #1072, opened 2024-07-17
Update Material type definitions: Add MaterialJSON format

Hoodgail:
Introduces the MaterialJSON format in Three.js, which extends the Material interface. The new format includes serializable properties such as color, roughness, metallic, map, normalMap, and many more. This change enables better JSON parsing and handling of material configurations for Three.js applications.

Confirmed: Types and interfaces have been updated in Object3D.d.ts and Material.d.ts. The Material class has also been updated to include toJSON methods that return MaterialJSON or MaterialJSONRoot objects based on the provided meta data.

Reference(s):
#1071
#1070
#426


Methuselah96 (approved):
Thanks!

PR #1078, opened 2024-07-18
WebGPURenderer: Build - three.webgpu.js

Methuselah96:
https://github.com/mrdoob/three.js/pull/28650

TODO

- [x] Rename examples-jsm
- [x] Go over other changes in PR
- [x] Fix local imports


Methuselah96:
The DT CI is failing, but hopefully that's just because three@0.167.0 is no released yet so there's a mismatch between the JS entry points and the declaration entry points.

PR #1122, opened 2024-07-23
fix: expose addNodeMaterial and createNodeMaterialFromType

0b5vr:
these functions are now exposed.

- See: https://github.com/mrdoob/three.js/blob/a00f79b32274975d5bd5c0f3d83ac8c31efec64d/src/nodes/materials/Materials.js#L3
- See: https://github.com/mrdoob/three.js/blob/a00f79b32274975d5bd5c0f3d83ac8c31efec64d/src/nodes/Nodes.js#L200


Methuselah96 (approved):
Thanks!

PR #1133, opened 2024-07-26
LightingModel: Refactor indirect lighting (2)

0b5vr:
`.ambientOcclusion` also integrated into `.indirect`

Sequel of 9724d5985707e7a81cae478e52fed0682a52d007

See: https://github.com/mrdoob/three.js/pull/28824


Methuselah96:
Are you sure? I still see `ambientOcclusion` right [here](https://github.com/mrdoob/three.js/blob/dev/src/nodes/core/LightingModel.js#L13).

Methuselah96:
Oh, I guess it just didn't get removed from the `LightingModel` class, even though it doesn't get called from `LightsNode` anymore. Can you create a PR on three.js to do that?

0b5vr:
I'm sorry I left this PR for a long time. I'm sure this PR is pretty stale; I will send you a PR again when it's still needed :bow:

PR #1135, opened 2024-07-29
ShaderNode: `If` return `StackNode`

linbingquan:
`If` should return `StackNode`.

Correct me if I'm wrong.


Methuselah96 (approved):
lgtm, thanks!

PR #1145, opened 2024-08-12
Remove loosely-typed event listener methods

### LINKED ISSUE #1143: EventDispatcher typing consistency
The `EventDispatcher` only has one method definition for  `dispatchEvent` which is strict.

https://github.com/three-types/three-ts-types/blob/c8cea806df55a23ff020819be3abb86c27a4adfe/types/three/src/core/EventDispatcher.d.ts#L84

<hr>
<br>

While there are multiple method definitions for `addEventListener` which make the typing a bit loose. 

https://github.com/three-types/three-ts-types/blob/c8cea806df55a23ff020819be3abb86c27a4adfe/types/three/src/core/EventDispatcher.d.ts#L52-L55

https://github.com/three-types/three-ts-types/blob/c8cea806df55a23ff020819be3abb86c27a4adfe/types/three/src/core/EventDispatcher.d.ts#L56

<hr>
<br>

I don't understand why  `addEventListener` is loose when it directly depends over `dispatchEvent` which is strict. Shouldn't the loosest implementation based over regular string be removed ? It would raise error when listening to none defined event and thus be more type safe 🤔
It will be especially useful when creating custom controls inheriting from `EventDispatcher`.

Methuselah96:
Fixes #1143


Methuselah96:
Waiting on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70279

s-rigaud:
In one of my projects, I ended up with a signature like this for a custom `TransformControls`, might be great to be added here I think:

```js
type ChangedEvents = {
  [K in `${string}-changed`]: any
}
type EventMap = THREE.Object3DEventMap & ChangedEvents & {
  change: unknown
  objectChange: unknown
  mouseDown: { mode: string }
  mouseUp: { mode: string }
}

class CustomTransformControls extends Object3D<EventMap> {
```

The special changed case is here to handle this special dispatch case

https://github.com/mrdoob/three.js/blob/4f067f0f4dc9e81f7fb3484962f2e973f71fab60/examples/jsm/controls/TransformControls.js#L89

PR #1146, opened 2024-08-14
StackNode: Rename `if` -> `If`, `else` -> `Else`, `elseif` -> `ElseIf`

linbingquan:
Related issue: https://github.com/mrdoob/three.js/pull/29064


Methuselah96 (approved):
Thanks!

PR #1223, opened 2024-09-09
#1222-DRACOLoader.d.ts-add-parse-function

### LINKED ISSUE #1222: Add interface for parse function to DRACOLoader.d.ts
1. parse function was called by load func(load from url, then parse)
2. In my case, my draco contents are transferred through other ways, load function is useless
3. Exposing parse function

Adamatoma:
1. parse function was called by load func(load from url, then parse)
2. In my case, my draco contents are transferred through other ways, load function is useless
3. Exposing parse function


Methuselah96 (approved):
Thanks!

Adamatoma:
> Thanks!

Btw, when will this pr be published?

Methuselah96:
It will be published at the end of the month with r169

PR #1286, opened 2024-10-16
fix: Now `element` from `uniformArray` accept both a `number` and `ShaderNodeObject`

Byloth:
https://discourse.threejs.org/t/how-to-access-an-uniformarray-element-using-a-node-object/72803


Methuselah96 (approved):
Thanks!

PR #1287, opened 2024-10-16
fix: Now `Else` node don't require a `Node` object anymore.

Byloth:
This is clearly some TYPO.

https://github.com/mrdoob/three.js/blob/1360b249b6fab26c92c2e467f57368faa69fd01b/src/nodes/core/StackNode.js#L63


Methuselah96 (approved):
Thanks!

PR #1290, opened 2024-10-20
fix: `MimeType` has been deprecated. `three` simply uses the `string` type.

Byloth:
This is something that clearly has changed over the time.

https://threejs.org/docs/#api/en/loaders/FileLoader.setMimeType


Methuselah96 (approved):
Thanks!

PR #1291, opened 2024-10-20
WebGLRendererParameters: "powerPreference" union type

satelllte:
Same change as https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70959/


Methuselah96 (approved):
Thanks! This will go out with `@types/three@0.170.0` at the end of the month.

PR #1336, opened 2024-11-04
TransformControls: Add scaleSnap

noname0310:
see https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/TransformControls.js#L106


Methuselah96 (approved):
Thanks!

PR #1375, opened 2024-11-12
feat(EventDispatcher): expose protected listeners

### LINKED ISSUE #1374: Expose EventDispatcher's _listeners as protected
This would allow users to seamlessly extend the behaviour of `EventDispatcher` when deriving from it.
As it stands, extending `EventDispatcher` does not allow the user to access the `_listeners` field without TypeScript errors, making code implementing utility functions worse. One common example would be clearing the listeners, which is not possible with the default `EventDispatcher` but is sometimes necessary.

HoloTheDrunk:
Resolves #1374 

Exposes the `_listeners` storage as a protected field so extending classes can access it.
The testsuite only worked once then never again, skipping with a fetch error on attw that I couldn't find anywhere online. I'm assuming this is due to the proxy at my work, but either way I can't test this myself at the moment. Seems to work though and it shouldn't be a breaking change.

### DIFF
--- a/types/three/src/core/EventDispatcher.d.ts
+++ b/types/three/src/core/EventDispatcher.d.ts
@@ -40,6 +40,27 @@ export type EventListener<TEventData, TEventType extends string, TTarget> = (
  */
 export class EventDispatcher<TEventMap extends {} = {}> {
     /**
+     * Maps event types to the list of their listeners.
+     * Only initialized when a listener is added.
+     * Event type entries are created when their first listener is added.
+     *
+     * @example
+     * ```typescript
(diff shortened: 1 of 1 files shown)


HoloTheDrunk:
Oof, seems like this is a much more significant change than I'd expected since nothing around listeners was strictly typed...
Might even be a breaking change considering that there was absolutely no type checking on event types or anything before.
I'll try to work through the errors one by one but help would be appreciated. This large node of missing typing being ignored for so long has led to a rather inconveniently large number of dependent code doing JS in TS.

PR #1377, opened 2024-11-23
fix(types): sync GLTFWriter and GLTFExporterPlugin types with impleme…

ybt-new:
These properties and methods are used in GLTF extensions but were missing from the type declarations, I've added them now.
- Add [extensionsUsed ](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/exporters/GLTFExporter.js#L506)property to GLTFWriter
- Add texture processing methods to GLTFWriter
  - [processTextureAsync](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/exporters/GLTFExporter.js#L1440)
  - [applyTextureTransform](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/exporters/GLTFExporter.js#L807)
- Fix writeMaterial type in GLTFExporterPlugin to match async implementation (writeMaterial -> [writeMaterialAsync](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/exporters/GLTFExporter.js#L1647))

### DIFF
--- a/types/three/examples/jsm/exporters/GLTFExporter.d.ts
+++ b/types/three/examples/jsm/exporters/GLTFExporter.d.ts
@@ -87,7 +87,7 @@ declare class GLTFExporter {
      */
     parse(
         input: Object3D | Object3D[],
-        onDone: (gltf: ArrayBuffer | { [key: string]: any }) => void,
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

PR #1386, opened 2024-12-01
add missing export

demike:
Else a deep import is necessary
from "three/src/textures/types"


Methuselah96:
Thanks, probably just better to remove the extra file altogether, done in https://github.com/three-types/three-ts-types/pull/1388.

PR #1398, opened 2024-12-12
loose events back

RodrigoHamuy:
#1145 introduced a breaking change on `EventDispatcher`, a class with an implementation that hasn't change in years.

The idea is great, but sadly it breaks a few things in `pmndrs` libraries as it makes backward compatibility really hard to keep.

I did try to fix it at `three-stdlib`, but I introduced more issues than solutions, also would prefer to avoid having to maintain in there a copy of `EventDispatcher` considering the real thing hasn't change in years.

I believe a much simpler solution that works for all is to just accept types to be a tiny bit looser.

With this we will still have types for `addEventListener` and so on, but is optional rather than required.

### DIFF
--- a/types/three/src/core/EventDispatcher.d.ts
+++ b/types/three/src/core/EventDispatcher.d.ts
@@ -53,6 +53,7 @@ export class EventDispatcher<TEventMap extends {} = {}> {
         type: T,
         listener: EventListener<TEventMap[T], T, this>,
     ): void;
+    addEventListener<T extends string>(type: T, listener: EventListener<{}, T, this>): void;
 
     /**
      * Checks if listener is added to an event type.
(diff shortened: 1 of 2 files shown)


Methuselah96:
Do you have a branch in `three-stdlib` where I can look at some of the difficulties you ran into?

RodrigoHamuy:
@Methuselah96 

You can check this Codesandbox with `three-stdlib@2.34.0` (before my first attempt at fixing the problem).

https://codesandbox.io/p/sandbox/gf3xzj

```ts
import { OrbitControls } from "three-stdlib";
import { PerspectiveCamera } from "three";

const orbit = new OrbitControls(new PerspectiveCamera());
orbit.addEventListener("change", (e) => {
  e.target.object;
});
```

Throws:

> Argument of type 'string' is not assignable to parameter of type 'never'.typescript(2345)

image

RodrigoHamuy:
@Methuselah96 if you want to check the rest of the conversation on the issue:

- issue: https://github.com/pmndrs/three-stdlib/issues/387
- first attempt at fixing it: https://github.com/pmndrs/three-stdlib/pull/388

PR #1463, opened 2024-12-22
ComputeNode: `workgroupSize` optional

### LINKED ISSUE #1448: TSL:  Missing overload for `Fn` and incorrect `Node` typing
## Description
Declaring a `Fn` function without a Node return type throws a TypeScript error:
```ts
const computeFn = Fn(() => {}); 
// Error: Argument of type '() => void' is not assignable to parameter of type '() => Node'.
```

This results in  a `computeFn` having an incorrect type, leading to further type errors when attempting to compute the function:
```ts
const computeNode = computeFn().compute();
// Error: Property 'compute' does not exist on type 'Node'.
```
It seems to only affects the type, the actual code runs without issues.

This issue was also mention on the Three.js discourse forum:
https://discourse.threejs.org/t/are-there-any-up-to-date-webgpu-169-examples/73378

I am willing to work on a pull request to fix this issue.

## Environment:
   "@types/three": "0.171"
    "three": "0.171"
    "typescript": "5.6.2"
    "vite": "6.0.1"
    
## Reproduction
A minimal reproducible example is available here:
https://github.com/stofi/three-types-tsl-fn-repro

linbingquan:
Fix: #1448

image

As the title says.

### DIFF
--- a/types/three/src/nodes/gpgpu/ComputeNode.d.ts
+++ b/types/three/src/nodes/gpgpu/ComputeNode.d.ts
@@ -19,7 +19,7 @@ export default class ComputeNode extends Node {
 export const compute: (
     node: NodeRepresentation,
     count: number,
-    workgroupSize: number[],
+    workgroupSize?: number[],
 ) => ShaderNodeObject<ComputeNode>;
 
 declare module "../tsl/TSLCore.js" {


Methuselah96 (approved):
Thanks!

PR #1493, opened 2025-01-09
ShaderNodeFn: improve DX for Layout

linbingquan:
As the title says.

### DIFF
--- a/types/three/src/nodes/tsl/TSLCore.d.ts
+++ b/types/three/src/nodes/tsl/TSLCore.d.ts
@@ -236,7 +236,7 @@ interface Layout {
     type: string;
     inputs: {
         name: string;
-        type: "bool" | "int" | "float" | "vec2" | "vec3" | "vec4";
+        type: string;
         qualifier?: "in" | "out" | "inout";
     }[];
 }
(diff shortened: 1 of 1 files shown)


types/three/src/nodes/tsl/TSLCore.d.ts
```diff
@@ -234,7 +234,11 @@ export function nodeImmutable<T>(
 interface Layout {
     name: string;
     type: string;
-    inputs: { name: string; type: string }[];
+    inputs: {
+        name: string;
+        type: "bool" | "int" | "float" | "vec2" | "vec3" | "vec4";
```
Methuselah96: Aren't there more possible types than this?

linbingquan: > Aren't there more possible types than this?

I don't remember why I wrote that...

Maybe use the type in the `NodeUtils.js -> getValueFromType` method, which should be more complete

```ts
type: "bool" | "int" | "uint" | "float" | "vec2" | "vec3" | "vec4" | "mat2" | "mat3" | "mat4" | "color" | "string" | "ArrayBuffer";
```

Methuselah96: I'm just going to change it back to `string` for now until I have time to look into it some more.

Methuselah96 (approved):

PR #1559, opened 2025-03-10
AnimationMixer: Add string as valid method input

s-rigaud:
Based over recent Jsdoc added, we can see that some methods accept a string as a valid parameter.

https://github.com/mrdoob/three.js/blob/dev/src/animation/AnimationMixer.js


Methuselah96 (approved):
Thanks!

PR #1560, opened 2025-03-11
TSLEncoder: add missing types

s-rigaud:
I added methods and types of TSLEncoder so we can get some type hinting for files such as https://github.com/mrdoob/three.js/blob/dev/examples/webgpu_shadertoy.html#L194.

I only added the bare minimum (no additional AST types) so it ends up with some any types but it is still better than before.

### DIFF
--- a/types/three/examples/jsm/transpiler/TSLEncoder.d.ts
+++ b/types/three/examples/jsm/transpiler/TSLEncoder.d.ts
@@ -1,7 +1,41 @@
-import { Program } from "./AST.js";
+import {
+    Conditional,
+    Expression,
+    For,
+    FunctionDeclaration,
+    Program,
+    Statement,
+    Ternary,
+    Uniform,
+    VariableDeclaration,
+    Varying,
+} from "./AST.js";
 
 export default class TSLEncoder {
     constructor();
 
+    tab: string;
+    imports: Set<string>;
+    global: Set<string>;
+    overloadings: Map<string, string>;
+    iife: boolean;
+    uniqueNames: boolean;
+    reference: boolean;
+
+    addImport(name: string): void;
+    emitUniform(node: Uniform): string;
+    emitExpression(node: Expression): string;
+    emitBody(body: Statement[]): string;
+    emitTernary(node: Ternary): string;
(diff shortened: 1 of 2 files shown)


Methuselah96:
Thanks! I'm not a huge fan of adding `any` types. How much work would it be to properly type those methods? If it's too much work, might be better to leave them out for now.

s-rigaud:
I updated the content of `AST.d.ts` to be able to remove all `any` types. The problem is that the AST is expecting the code to be well formatted to return the according type using the `parseExpressionFromTokens` function but the return type of the function is really wide so I ended up with a lot of generic `Statement` types.

types/three/examples/jsm/transpiler/TSLEncoder.d.ts
```diff
@@ -1,7 +1,30 @@
...
+    global: Set<string>;
+    overloadings: Map<string, string>;
+    iife: boolean;
+    uniqueNames: boolean;
+    reference: boolean;
+
+    addImport(name: string): void;
+    emitUniform(node: Statement): string;
+    emitExpression(node: Statement): string;
+    emitBody(body: Statement): string;
```
Methuselah96: This one is an array, right?

```suggestion
    emitBody(body: Statement[]): string;
```

types/three/examples/jsm/transpiler/TSLEncoder.d.ts
```diff
@@ -1,7 +1,30 @@
...
+    tab: string;
+    imports: Set<string>;
+    global: Set<string>;
+    overloadings: Map<string, string>;
+    iife: boolean;
+    uniqueNames: boolean;
+    reference: boolean;
+
+    addImport(name: string): void;
+    emitUniform(node: Statement): string;
```
Methuselah96: I'm still a little confused why we can't make these parameter types more specific. Doesn't this method assume the parameter is a `Uniform`, for example?

s-rigaud: You're right, I updated the definition 👍

Methuselah96 (approved):
Thanks!

PR #1579, opened 2025-03-19
Adds typed uniforms to ShaderMaterial and RawShaderMaterial

haxiomic:
I've been using this for years in my own projects where I use custom shaders often. Typing uniforms makes working with these classes much nicer

```ts
let shader = new ShaderMaterial({ uniforms: { example: new Uniform(3), ... }, ... });

// typed correctly!
shader.uniforms.example.value = 4
```

This does not change the behaviour of existing code because uniform types default to `any`

### DIFF
--- a/types/three/src/materials/RawShaderMaterial.d.ts
+++ b/types/three/src/materials/RawShaderMaterial.d.ts
@@ -1,7 +1,9 @@
 import { ShaderMaterial, ShaderMaterialParameters } from "./ShaderMaterial.js";
 
-export class RawShaderMaterial extends ShaderMaterial {
-    constructor(parameters?: ShaderMaterialParameters);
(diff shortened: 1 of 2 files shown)


haxiomic:
I'll add Defines types before this is complete

LePhenix47:
This should get reviewed and merged. Right now I hand-write uniform types and boilerplate helper types just to get autocomplete on uniforms.x.value. 

Generic ShaderMaterial/RawShaderMaterial would kill most of that code. Example of what I'm working around:

https://github.com/LePhenix47/ThreeJS_test/blob/shader-lighting/src/modules/World/ShadingGroup.ts
https://github.com/LePhenix47/ThreeJS_test/blob/shader-lighting/src/modules/World/types/uniforms.ts

PR #1586, opened 2025-03-25
LutPass: no params is a valid constructor

s-rigaud:
See here : https://github.com/mrdoob/three.js/blob/35b0fbc461520006769514eece8c3b371a79257c/examples/jsm/postprocessing/LUTPass.js#L76

We can use no arguments in the constructor and still instantiate a LutPass.


Methuselah96 (approved):
Thanks!

PR #1631, opened 2025-05-06
Add missing CapsuleGeometry parameter

mkeblx:
Missing parameter.

See:
https://github.com/mrdoob/three.js/blob/r176/src/geometries/CapsuleGeometry.js#L46


Methuselah96 (approved):
Thanks!

PR #1634, opened 2025-05-19
fix: Add morphAttributes to `Partial`

0b5vr:
The type of `morphAttributes` is currently `Record<"position" | "normal" | "color", Array<BufferAttribute | InterleavedBufferAttribute>>`, and that means every one of these three properties must be defined, which is not true. [The class definition also violates it](https://github.com/mrdoob/three.js/blob/r176/src/core/BufferGeometry.js#L131).

The previous change is made by this commit in #1558: https://github.com/three-types/three-ts-types/pull/1558/commits/fae6c44b22ab8de9aa571b95433efdd5b08db658


0b5vr:
seems tests are failing, I don't know how to fix this. how do I find the instruction?

Methuselah96 (approved):
Thanks! I fixed the test.

PR #1635, opened 2025-05-20
Fix Keyframe subclass constructor arguments

### LINKED ISSUE #1633: The argument array type should be `ArrayLike` instead of `Array`
### Description

The argument array type of KeyframeTracks has been changed from `ArrayLike` to `Array`, and that prevents giving a `Float32Array` to the constructor.

The change was made in this commit in #1578:

https://github.com/three-types/three-ts-types/blob/5e1ece2b66e61742fb390eee93b4d7fa3958a0de/types/three/src/animation/tracks/QuaternionKeyframeTrack.d.ts#L16

That should be fixed since the Three.js codebase is intended to use the given array instance as is if the given array type is same as the one it internally uses.

https://github.com/mrdoob/three.js/blob/r176/src/animation/KeyframeTrack.js#L45
https://github.com/mrdoob/three.js/blob/r176/src/animation/AnimationUtils.js#L11

### Workaround

Change the type of the argment to `ArrayLike` in the constructor of the class.

Patch can be done by using [patch-package](https://www.npmjs.com/package/patch-package).

Another workaround is to use `@ts-expect-error`.

### Expected measurements

I assume we have to change the jsdoc type of the Three.js side, since it generates the types from it.

Methuselah96:
Fixes #1633


0b5vr:
Thank you!

PR #1670, opened 2025-06-03
Update AxesHelper generics

junepil:
### Examples
```ts
const axes = new THREE.AxesHelper();
axes.renderOrder = 1;
// missing in type definition, because the mateiral type is Material | Material []
axes.material.depthTest = false; 
```
### Changes
Narrows down the [AxesHelper](https://github.com/mrdoob/three.js/blob/dev/src/helpers/AxesHelper.js)'s type so the type correction works accurately. AxesHelper only uses `BufferGeomtry` and `LineBasicMaterial` inside the constructor. So I believe there are no reasons that the `Material []` type should be applied.

image


Methuselah96 (approved):
Thanks!

PR #1687, opened 2025-06-14
Fix Typo in SizeVector3Uninitialized Type

zeevick10:
**Description:**  
This pull request includes the following changes:
- Corrected a typo in the author name from "Brower" to "Browser" in the lil-gui module type definitions.
- Refactored the `SizeVector3Uninitialized` type in `Textures.d.ts` for improved clarity and consistency.
- Updated the `getSize` method signature to use the new `SizeVector3Uninitialized` type.


types/three/examples/jsm/libs/lil-gui.module.min.d.ts
```diff
@@ -324,7 +324,7 @@ export class ColorController<T = Record<string, unknown>, K extends keyof T = ke
  * lil-gui
  * https://lil-gui.georgealways.com
  * @version 0.17.0
- * @author George Michael Brower
+ * @author George Michael Browser
```
Methuselah96: Are you sure this change is correct? Looks like Brower in the [lil-gui package.json](https://github.com/georgealways/lil-gui/blob/master/package.json#L5).

Methuselah96 (approved):
Thanks!

PR #1707, opened 2025-06-22
Strongly type `StructNode`

### LINKED ISSUE #1632: get method missing from ShaderNodeObject<Node> in Typescript definition
When using structs with TSL in a typescript based project, the compiler complains that `Property 'get' does not exist on type ShaderNodeObject<Node>` (or `ShaderNodeObject<StructNode>`) when I try to access the members of a struct. I have to cast to any and it will work. Obviously, the javascript is fine, but something is missing in the typescript definition.

As an example:
```
const BoundingBox = struct(
    {
        start: {type:"vec3"},
        end: {type: "vec3"}
    },
    "BoundingBox"
);

const boundingBoxArray = instancedArray(particleCount, BoundingBox);
const boundingBox =  boundingBoxArray.element(instanceIndex.toInt().div(atomCount));

const start = boundingBox.get("start"); // typescript error
const start = (boundingBox as any).get("start"); // works
```

Methuselah96:


Methuselah96:
Superseded by https://github.com/three-types/three-ts-types/pull/1856, but it will still be nice for this to be strongly typed...

PR #1726, opened 2025-07-30
Update StructNode.d.ts

### LINKED ISSUE #1632: get method missing from ShaderNodeObject<Node> in Typescript definition

novogrammer:
## Summary

- **Add** `get()` method to `Swizzable`  
- **Add** support for `Record<string, Node>` overload in `Struct`  
- **Remove** invalid `(values: Node[])` overload from `Struct`  
- **Add** unit tests for `StructNode.d.ts`

## Related Issue

Closes https://github.com/three-types/three-ts-types/issues/1632

## See Also

https://github.com/mrdoob/three.js/pull/30394

### DIFF
--- a/types/three/test/unit/src/nodes/core/StructNode.ts
+++ b/types/three/test/unit/src/nodes/core/StructNode.ts
@@ -1,0 +1,69 @@
+
+import { Fn, vec3, struct, float } from "three/tsl";
+
+
+const Complex = struct({
+    real:'float',
+    imaginary:'float',
+});
+
+const multiply = Fn(([a,b]:[a:ReturnType<typeof Complex>,b:ReturnType<typeof Complex>])=>{
+    const ar=a.get("real");
+    const ai=a.get("imaginary");
+    const br=b.get("real");
+    const bi=b.get("imaginary");
+    return Complex({
+        real: ar.mul(br).sub(ai.mul(bi)),
+        imaginary: ar.mul(bi).add(ai.mul(br))
+    });
+})
+
+const complexA = Complex(float(2),float(1));
+const complexB = Complex({
+    real:float(2),
+    imaginary:float(-1),
+});
+const complexC = multiply(complexA,complexB);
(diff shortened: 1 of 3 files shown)


types/three/test/unit/src/nodes/core/StructNode.ts
```diff
@@ -0,0 +1,69 @@
...
+const multiply = Fn(([a,b]:[a:ReturnType<typeof Complex>,b:ReturnType<typeof Complex>])=>{
+    const ar=a.get("real");
+    const ai=a.get("imaginary");
+    const br=b.get("real");
+    const bi=b.get("imaginary");
+    return Complex({
+        real: ar.mul(br).sub(ai.mul(bi)),
+        imaginary: ar.mul(bi).add(ai.mul(br))
+    });
+})
```
Methuselah96, line 19: Thanks for the PR! Unfortunately I don't think this is strict enough. I think ideally we would want `a.get("invalid")` to produce a type-error, but it looks like it currently allows any string key.

novogrammer: I’ll close this PR because it’s hard to keep it consistent with the other features.
Thank you for your time and feedback!

PR #1803, opened 2025-09-01
fix: Add branches for `JoinNode` in `mat*` node functions

sxxov:
`mat*` functions currently only accept numbers in the constructor to create `ConstNode`s of matrices. however, since they use the same code path as `vec*`, & [core source](https://github.com/mrdoob/three.js/blob/27151c8325d1dba520d4abfb5a2e1077dd59de22/src/nodes/functions/BSDF/LTC.js#L92) shows that they can accept nodes, this PR adds the same signature as `vec*`s to align them.


Methuselah96:
Superseded by https://github.com/three-types/three-ts-types/pull/1855

PR #1805, opened 2025-09-02
fix: Add `assign`, `toVarIntent`, & `get` methods to `Node`

### LINKED ISSUE #1804: `Node` is missing dynamically added prototype methods
in [`TSLCore.js`](https://github.com/mrdoob/three.js/blob/27151c8325d1dba520d4abfb5a2e1077dd59de22/src/nodes/tsl/TSLCore.js#L58C1-L95C1) (previously defined using proxies, now tacked on the prototype), `Node` is defined with 3 extra properties--`assign`, `toVarIntent`, `get`.

it seems to not be present in [`Node.d.ts`](https://github.com/three-types/three-ts-types/blob/master/types/three/src/nodes/core/Node.d.ts) or in [`TSLCore.d.ts`](https://github.com/three-types/three-ts-types/blob/master/types/three/src/nodes/tsl/TSLCore.d.ts), which means trying to do `struct.get('prop')` results in a type error.

usage example for `get` can be seen in the PR for `struct` here: https://github.com/mrdoob/three.js/pull/30394

sxxov:
fixes: https://github.com/three-types/three-ts-types/issues/1804

going forwards it seems like upstream is ditching the proxy way of adding dynamic properties & tacking them on the prototype instead. this means the types should probably reflect that & apply the swizzling & 0-31 index accesses onto `Node` itself instead of `ShaderNodeObject`

### DIFF
--- a/types/three/src/nodes/core/Node.d.ts
+++ b/types/three/src/nodes/core/Node.d.ts
@@ -358,7 +358,7 @@ declare class Node extends EventDispatcher<{
      * @return {Object} The serialized node.
      */
     toJSON(meta?: NodeJSONMeta | string): NodeJSONOutputData;
-    assign(...params: Parameters<typeof assign>): this;
+    assign(sourceNode: Node | number): this;
     toVarIntent(): this;
     get(value: string): MemberNode;
 }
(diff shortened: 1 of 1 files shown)


Methuselah96:
Superseded by https://github.com/three-types/three-ts-types/pull/1856 (and https://github.com/three-types/three-ts-types/pull/1839)

PR #1810, opened 2025-09-04
add update to LoopNodeObjectParameter

s-rigaud:
Loop accept an update VarNode https://github.com/mrdoob/three.js/blob/61f467b53d26e24ca36918e6bdc27f0f3db3f697/examples/jsm/tsl/utils/Raymarching.js#L62

### DIFF
--- a/types/three/src/nodes/utils/LoopNode.d.ts
+++ b/types/three/src/nodes/utils/LoopNode.d.ts
@@ -1,5 +1,6 @@
 import Node from "../core/Node.js";
 import NodeBuilder from "../core/NodeBuilder.js";
+import { VarNode } from "../Nodes.js";
 import { ShaderNodeObject } from "../tsl/TSLCore.js";
 
 interface LoopNodeObjectParameter {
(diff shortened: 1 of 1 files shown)


cmhhelgeson:
It should be able to take a VarNode and a number and a string right? Those are all represented as types in the LoopNode generator function?

```ts
// src/nodes/utils/LoopNode.js
switch ( typeof update ) {
	case 'function':
		const flow = builder.flowStagesNode( properties.updateNode, 'void' );
		const snippet = flow.code.replace( /\t|;/g, '' );
		updateSnippet = snippet;
		break;
	case 'number':
		updateSnippet = name + ' ' + deltaOperator() + ' ' + builder.generateConst( type, update );
		break;
	case 'string':
		updateSnippet = name + ' ' + update;
		break;
	default:
		if ( update.isNode ) {
			updateSnippet = name + ' ' + deltaOperator() + ' ' + update.build( builder );
		} else {
			error( 'TSL: \'Loop( { update: ... } )\' is not a function, string or number.' );
			updateSnippet = 'break /* invalid update */';
		}
}
```

String Example that Three.js accepts:

```ts
Loop( { start: uint( 2 ), end: uint( workgroupSize * 2 ), type: 'uint', condition: '<=', update: '<<= 1' }, () => {
```

s-rigaud:
True, I just didn't encounter the other types myself before

Methuselah96 (approved):

PR #1814, opened 2025-09-18
TSL: Allow TextureNode to accept Texture or TextureNode

huntercaron:
The `texture()` and `textureLoad()` functions in Three.js TSL accept TextureNode as well as Texture. The inline type for this can be seen [here](https://github.com/mrdoob/three.js/blob/3c32743c1e5355f43d9e8319086caf92a0b19955/src/nodes/accessors/TextureNode.js#L822). This has been the case since r177 and the change can be found in [this PR](https://github.com/mrdoob/three.js/pull/31190).


huntercaron:
Following branch naming convention, re-opened here https://github.com/three-types/three-ts-types/pull/1815

PR #1816, opened 2025-09-21
Vector3: Fix parameter names

xZise:
Use the names from the original implementation to also match the documentation.


xZise:
As requested in [DefinitelyTyped #73624](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73624).

Methuselah96 (approved):

Issue #1837, opened 2025-10-10
error TS2552: Cannot find name 'GPUTexture


roaldjap:
```ts
Error: node_modules/@types/three/src/textures/ExternalTexture.d.ts:4:35 - error TS2552: Cannot find name 'GPUTexture'. Did you mean 'Texture'?   

4     sourceTexture: WebGLTexture | GPUTexture | null;
                                    ~~~~~~~~~~


Error: node_modules/@types/three/src/textures/ExternalTexture.d.ts:8:48 - error TS2552: Cannot find name 'GPUTexture'. Did you mean 'Texture'?   

8     constructor(sourceTexture?: WebGLTexture | GPUTexture | null);
```

After installing. this is what I get - it seems GPUTexture wrongly defined here.

Methuselah96:
@roaldjap This should be fixed in `@types/three@0.181.0`, can you give it a try when you get a chance?

roaldjap:
@Methuselah96 it didnt work. Just reverted back on older version instead.

Methuselah96:
Okay, I can't reproduce the issue, feel free to post an update if you can provide a reproducible example repo

PR #1838, opened 2025-10-10
Replace GPUTexture to Texture

roaldjap:
Fixing issue for: https://github.com/three-types/three-ts-types/issues/1837


Methuselah96:
The current type is intentional, https://github.com/three-types/three-ts-types/pull/1821 might fix your issue when it is released with `@types/three@0.181.0`

PR #1883, opened 2025-11-06
Fix conversion methods capitalization

### LINKED ISSUE #1882: r181 Mismatch API names, `toUVec2`(three) vs `toUvec2`(three-ts-types) etc.
three.js wiki: https://github.com/mrdoob/three.js/wiki/Three.js-Shading-Language#conversions

three.js source: https://github.com/mrdoob/three.js/blob/r181/src/nodes/tsl/TSLCore.js#L1177-L1196

three-ts-types: https://github.com/three-types/three-ts-types/blob/f495c22f37ef43c070237379b0715d79d85b3da7/types/three/src/nodes/tsl/TSLCore.d.ts#L467-L527

Methuselah96:
Fix https://github.com/three-types/three-ts-types/issues/1882

### DIFF
--- a/types/three/src/nodes/tsl/TSLCore.d.ts
+++ b/types/three/src/nodes/tsl/TSLCore.d.ts
@@ -483,38 +483,38 @@ declare module "../Nodes.js" {
         toVec2: () => Node;
         toVec2Assign: () => this;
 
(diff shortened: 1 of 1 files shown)


ycw:
Thanks.

PR #1930, opened 2025-12-02
CubeTexture: make image type generic

Mirei3D:
CubeTexture.images can hold a `DataTexture[]`, for example when [loading a HDR with HDRCubeTextureLoader](https://github.com/mrdoob/three.js/blob/65bfbd8e51db937ecd5d4be7f2d93319f6e5ee97/examples/jsm/loaders/HDRCubeTextureLoader.js#L113-L122) or an array of `{width, height, depth}` objects when used as part of a [WebGLCubeRenderTarget](https://github.com/mrdoob/three.js/blob/7703433fca8aa2af2ffefe1c83dd17139298f2ce/src/renderers/WebGLCubeRenderTarget.js#L36-L37)

Untested at the time of creating this PR as I couldn't get `npm run test` to run locally.
Let me know if there is anything I should change.


Methuselah96 (approved):
Thanks!

PR #1969, opened 2026-01-13
fix: Make element from uniformArray accept both a number and Node

thelazylamaGit:
Reversion back to an old fix https://github.com/three-types/three-ts-types/pull/1286. Forum post from 2024 discussing the issue: https://discourse.threejs.org/t/how-to-access-an-uniformarray-element-using-a-node-object/72803/6


Methuselah96 (changes requested):
Thanks! `NodeRepresentation` no longer exists, can you use `Node | number` instead?

Methuselah96 (approved):

PR #2045, opened 2026-02-20
Fix ComputeBuiltinNode typings for vector builtins (uvec3 swizzles)

thelazylamaGit:
Gives proper typing for globalId, localId, etc, allowing the use of builtin `.x/.y/.z`


Methuselah96 (approved):
Thanks!

PR #2051, opened 2026-02-21
TSL: Improve types for Fn

Methuselah96:


Methuselah96:
`NodeBuilder.format` is responsible for conversions

Methuselah96:
Might be helpful for converting object to tuple: https://stackoverflow.com/questions/52855145/typescript-object-type-to-array-type-tuple

PR #2061, opened 2026-02-27
Fix WebXRController export

potrepka:
`WebXRController` exports in Three.d.ts are incorrect (type-only) and are overshadowing exports in Three.Core.d.ts.

Fix: Remove the `WebXRController` exports from Three.d.ts.


Methuselah96 (approved):
Thanks! It looks like `WebXRController` is exported as a value from `Three.Core.js`, but the rest of the exports are still type-only exports so I've updated the export in `Three.Core.d.ts` to reflect that.

PR #2097, opened 2026-03-27
TSL: Make Switch Case return void

thelazylamaGit:
Currently when doing a Switch statement in TSL such as

```ts
Switch( 0 ).Case( 0, () => {
    col.assign( color( 1, 0, 0 ) );
})
```

typescript complains

```error
Argument of type '() => void' is not assignable to parameter of type 'Node'. Type '() => void' is not assignable to type 'NodeClass'.
```
Simple fix is to change Case type in StackNode.d.ts from
```ts
Case(...params: Node[]): this;
```
to
```ts
Case(...params: [...Node[], () => void]): this;
```

### DIFF
--- a/types/three/src/nodes/core/StackNode.d.ts
+++ b/types/three/src/nodes/core/StackNode.d.ts
@@ -15,7 +15,7 @@ declare class StackNode extends Node {
 
     Switch(expression: Node): this;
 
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

PR #2105, opened 2026-04-03
LDrawLoader: Fix materialLibrary property spelling

NicklasKnell:
Just stumbled across this type mismatch.
https://github.com/mrdoob/three.js/blob/master/examples/jsm/loaders/LDrawLoader.js#L1794

Wasn't really sure on how to properly reflect this change in the tests, so i am very much open for feedback.


Methuselah96:
Thanks! Since it's just a typo that's unlikely to regress, I just left the test changes out.

PR #2112, opened 2026-04-11
Always add EventMap type parameter for classes derived from EventDispatcher

shotamatsuda:
This is a long-standing issue with EventDispatcher to me. EventDispatcher only accepts events that are declared in EventMap, but some derived classes don't have type parameter or overloadable interface for it.

This can be a breaking change when the users infer the type parameters, though the impact should be minimal.


Methuselah96 (approved):
Thanks!

PR #2115, opened 2026-04-11
Add missing types to UniformNode

shotamatsuda:
This PR adds missing types to UniformNode. They are undocumented but commonly used throughout TSL code in Three.js.

1. Factory function with a type name:

```ts
const node = uniform('vec3') // UniformNode<'vec3', Vector3>
```

2. `this` as the last parameter of callbacks passed to onUpdate-related functions.

```ts
declare const matrix: Matrix4
const node = uniform('vec3').onRenderUpdate(
  (_, { value }) => value.setFromMatrixPosition(matrix)
)
```


shotamatsuda:
@Methuselah96 Just to confirm, am I correct that the JSDoc tests are intended to detect suspicious changes and potential incompatibilities made by changes in Three.js, or should I update JSDocs as part of PR?

Methuselah96:
I will update the JSDoc patch don't worry about that. We do generate some types from the JSDoc with patches on top of that, but the process for updating those is not well documented, so I'll just update those myself.

Methuselah96 (approved):
Thanks, I made some small changes and added some tests

PR #2128, opened 2026-04-23
AtomicFunctionNode: Fix return types

thelazylamaGit:
When working with atomic nodes such as 
```js
const count = float(atomicLoad(trailCounts.element(idx)))
```
typescript would complain because `AtomicFunctionNode` wasn't typed `<'uint'>` properly

```
No overload matches this call.
Overload 1 of 2, '(value?: number | undefined): VarNode<"float", ConstNode<"float", number>>', gave the following error. Argument of type 'AtomicFunctionNode' is not assignable to parameter of type 'number'.
```
this PR fixes the issue having it return
```js
atomicLoad<"uint">(pointerNode: Node<"uint">): AtomicFunctionNode<"uint">
```
instead of 
```js
atomicLoad(pointerNode: Node): AtomicFunctionNode
```

### DIFF
--- a/types/three/src/nodes/gpgpu/AtomicFunctionNode.d.ts
+++ b/types/three/src/nodes/gpgpu/AtomicFunctionNode.d.ts
@@ -30,8 +30,7 @@ declare class AtomicFunctionNodeClass extends Node {
     static ATOMIC_XOR: "atomicXor";
 }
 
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

PR #2129, opened 2026-04-28
[three] Revert static is(Vector|Matrix*) changes.

bzztbomb:
Most ThreeJS classes have a is* property on the prototype.  This allows for typechecks without using instanceof.  Recently the vector and matrix classes change to initialize this property using static initialization block. (PR here: https://github.com/mrdoob/three.js/pull/33140)

The types were updated to be 'static readonly isVector3: boolean' which means one can only access isVector3 like this: Vector3.isVector3 which is not useful. While the initialization of the property changed, the shape of the class did not.

This commit reverts the type update and restores the previous functionality.

Thanks!

### DIFF
--- a/types/three/src/math/Matrix2.d.ts
+++ b/types/three/src/math/Matrix2.d.ts
@@ -12,7 +12,7 @@ export type Matrix2Tuple = [
  * const m = new Matrix2();
  */
 export class Matrix2 {
-    static readonly isMatrix2: boolean;
+    readonly isMatrix2: true;
 
     /**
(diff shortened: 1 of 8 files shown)


bzztbomb:
Fixed the bad path in tsconfig, sorry about that!

Methuselah96 (approved):
Thanks!

Methuselah96:
Created https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74932 to get this released as a bug fix on r184

bzztbomb:
Thank you! Thanks for maintaining these types!

PR #2135, opened 2026-05-05
fix: make smoothstep and step available as chained methods on vec nodes

### LINKED ISSUE #2134: `smoothstep` and `step` not available as chained methods on vec2/vec3/vec4 nodes
`smoothstep` and `step` are not typed as chained methods on vector nodes, even though they work correctly at runtime and similar methods like `clamp` are properly typed.

To reproduce:
```ts
import { vec3 } from "three/tsl";

const v = vec3(1, 2, 3);
v.add(1).smoothstep(0.1, 0.5);  // ts error: Property 'smoothstep' does not exist on type 'Node<"vec3">'.
v.add(1).step(0.5);  // similar error
v.add(1).clamp(0, 1);  // ok
```

HannahLilyW:
Fixes #2134 

`smoothstep` and `step` were only typed as chained methods on `Node<"float">`, despite working correctly at runtime on all vector types. This adds missing extensions for `vec2`, `vec3`, and `vec4`, and also adds vector overloads to the standalone `smoothstep`, `smoothstepElement`, `step`, and `stepElement` functions.

The fix follows the same pattern already used by `clamp`.

Tests added in `tsl-testing/node-type.test.ts`.

### DIFF
--- a/tsl-testing/node-type.test.ts
+++ b/tsl-testing/node-type.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'vitest';
-import { Fn, vec2, vec3, vec4 } from 'three/tsl';
+import { Fn, smoothstep, smoothstepElement, step, stepElement, vec2, vec3, vec4 } from 'three/tsl';
 import * as THREE from 'three/webgpu';
 
 const renderer = new THREE.WebGPURenderer();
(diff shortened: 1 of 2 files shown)


Methuselah96 (approved):
Thanks!

PR #2136, opened 2026-05-06
Timer: Fix dispose return type

eunjiyun:
## Summary

- Aligns `Timer.dispose()` with the JavaScript implementation.
- Keeps the change limited to the declaration file.

## Evidence

- JS implementation: `src/core/Timer.js:140`
- Declaration: `types/three/src/core/Timer.d.ts:66`
- Mismatch: the declaration exposed a chainable return value that the implementation does not return.

## Check

- Not run: dependencies were not installed locally.
- Checked the declaration diff manually.

### DIFF
--- a/types/three/src/core/Timer.d.ts
+++ b/types/three/src/core/Timer.d.ts
@@ -63,7 +63,7 @@ declare class Timer {
     /**
      * Can be used to free all internal resources. Usually called when the timer instance isn't required anymore.
      */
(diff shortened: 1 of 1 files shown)


Methuselah96 (approved):
Thanks!

Methuselah96:
CI is not working for some reason, probably unrelated to this PR. Will merge once that is fixed.

PR #2145, opened 2026-05-07
TSL: Make `ArrayNode` Generic

thelazylamaGit:
`ArrayNode` was returning an untyped `Node[]` so wasn't working with the new `Node<TNodeType>` system. This PR makes `ArrayNode` generic so you can use it with the correct type such as `ArrayNode<'vec3'>`.

Likewise, makes `.element(i)` return the correctly typed `ArrayElementNode<TNodeType>`.


Methuselah96 (approved):
Thanks! I cleaned up some of the updates, let me know if I broke anything.

PR #2150, opened 2026-05-12
StorageTextureNode: Tighten types & add store function

thelazylamaGit:
Modifies StorageTextureNode.d.ts to type value as `StorageTexture` instead of `Texture` to better match the three.js definitions and also adds new type declaration for `.store()` function added in this PR https://github.com/mrdoob/three.js/pull/33544


thelazylamaGit:
test-examples seems to be failing because its inferring the js `Fn(({ storageTexture }) => ...)` as `Texture`, not `StorageTexture`.

`StorageTexture | StorageTextureNode` better matches the JSDoc & should be more accurate here, but `Texture | StorageTextureNode` avoids this inference error and should still work anyways. I'll leave choice up to whatever maintainer thinks is better.

Methuselah96 (approved):
Thanks!

PR #2204, opened 2026-06-08
WebGLState: Add getParameter and pixelStorei

mrxz:
With r184 the `WebGLState` class gained two new methods as part of https://github.com/mrdoob/three.js/pull/33227. These are accessible through the renderer (`renderer.state`) and are thus part of the public API. This PR adds both `pixelStorei` and `getParameter` to the `WebGLState.d.ts` file.

Both methods are effectively wrappers for the corresponding ones in `WebGLRenderingContext`. I've based the parameter and return types on the ones used in `lib.dom.d.ts` but kept the names as they appear in the Three.js source code.


Methuselah96 (approved):
Thanks!

PR #2205, opened 2026-06-08
feat(OrbitControls): add generic camera type parameter

woo-cie:
Make `OrbitControls` generic over its camera type so that `controls.object` retains the concrete camera type passed to the constructor instead of widening to `Camera`.

## What changed

- `Controls` base class: added a `TObject extends Object3D = Object3D` type parameter; `object` is now typed `TObject` and the constructor accepts `TObject`.
- `OrbitControls`: added a `TCamera extends Camera = Camera` type parameter and forwards it to `Controls<OrbitControlsEventMap, TCamera>`; the constructor accepts `TCamera`.
- Added `test/unit/examples/jsm/controls/OrbitControls.ts` with `$ExpectType` tests.

```ts
const c0 = new OrbitControls(new THREE.PerspectiveCamera());
c0.object; // PerspectiveCamera (was: Camera)

const c3: OrbitControls = new OrbitControls(new THREE.PerspectiveCamera());
c3.object; // Camera (default type argument — backward compatible)
```

All type parameters have defaults, so existing usages (including `MapControls extends OrbitControls` and `const c: OrbitControls`) remain source-compatible.

---

Ported from DefinitelyTyped PR [DefinitelyTyped/DefinitelyTyped#75077](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/75077).

### DIFF
--- a/types/three/examples/jsm/controls/OrbitControls.d.ts
+++ b/types/three/examples/jsm/controls/OrbitControls.d.ts
@@ -20,7 +20,7 @@ export interface OrbitControlsEventMap {
 /**
  * Orbit controls allow the camera to orbit around a target.
  */
-declare class OrbitControls extends Controls<OrbitControlsEventMap> {
+declare class OrbitControls<TCamera extends Camera = Camera> extends Controls<OrbitControlsEventMap, TCamera> {
     /**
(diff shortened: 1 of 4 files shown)


Methuselah96 (approved):
Thanks!

PR #2242, opened 2026-06-30
fix (GLTFLoader): add `primitives` to `GLTFReference`

0b5vr:
GLTFLoader can put `primitives` to `parser.associations` to mesh primitives

See: https://github.com/mrdoob/three.js/blob/r184/examples/jsm/loaders/GLTFLoader.js#L3909-L3912


Methuselah96 (approved):
Thanks!

